PR #10, opened 2020-04-21
Provision logging secrets per account

kenoir:
This change provisions logging secrets in client accounts in order to simplify cross account permissions and to decentralise access to secrets.

Part of https://github.com/wellcomecollection/platform/issues/4435


shared/modules/secret/main.tf
```diff
@@ -0,0 +1,8 @@
+resource "aws_secretsmanager_secret" "secret" {
+  name = var.name
+}
+
+resource "aws_secretsmanager_secret_version" "example" {
```
alexwlchan, line 5: "example" is the wrong name for this resource

shared/modules/secrets/distributed/README.md
```diff
@@ -0,0 +1,26 @@
+# Distributed secrets
+
+This module copies secrets from the platform account into client accounts as required.
```
alexwlchan, line 3: It copies them _and keeps them in sync_, I think?

kenoir:
> You could make this a bit more generic by allowing the caller to pass in an "aws.src" and "aws.dst"

I tried to follow the explicit provide rules but I couldn't actually get this to work (as per the terraform docs, which I think you are working from here).

PR #12, opened 2020-04-24
Redirect alpha site to respective github repo

jamesgorrie:
Part of https://github.com/wellcomecollection/wellcomecollection.org/issues/4088


kenoir:
Does this apply cleanly?

jamesgorrie:
> Does this apply cleanly?

The state was out of whack, it now isn't, and applies like a dream.

PR #17, opened 2020-04-29
Add an explicit document type for the fluentbit container

alexwlchan:
This wasn't caught when setting up the logging because we didn't have an index template.

The index templates require that the document type is `_doc` – if it’s possible to change that, it's very well hidden. Easier to lean into the ES way of doing things, and ditch the custom document type.

This change is published and running in the storage staging.

### DIFF
--- a/containers/fluentbit/extra.conf
+++ b/containers/fluentbit/extra.conf
@@ -13,3 +13,10 @@
     Logstash_Format On
     Logstash_Prefix firelens
     tls             On
+
+    # By default, Fluentbit sends the document type "flb_type".  Elasticsearch
+    # is gradually deprecating multiple types per index, and the index template
+    # for our firelens indexes specifies the type "_doc".
+    #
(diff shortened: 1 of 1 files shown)


containers/fluentbit/extra.conf
```diff
@@ -12,4 +12,11 @@
     HTTP_Passwd     ${ES_PASS}
     Logstash_Format On
     Logstash_Prefix firelens
-    tls             On
\ No newline at end of file
+    tls             On
+
+    # By default, Fluentbit sends the document type "flb_type".  Elasticsearch
+    # is gradually deprecating multiple types per index, and the index template
```
jamesgorrie, line 18: [As of V8](https://www.elastic.co/guide/en/elasticsearch/reference/current/removal-of-types.html), which we will hopefully be able to upgrade to nicely because of paying attention to this.
💯

PR #20, opened 2020-06-01
Configure Loris's nginx to use the uwsgi protocol over a unix socket

jamieparkinson:


jamieparkinson:
Does anyone know why https://github.com/wellcomecollection/platform-infrastructure/pull/20/files#diff-88c5b05219384771c339ea0622eda155R26 is hardcoded to `iiif.wellcomecollection.org`? Is there any reason we can't change that to `$server_name`?

PR #21, opened 2020-06-02
Adds new Route53 management role

kenoir:
Hosted zone records for wellcomecollection.org are moving to a new account, access to that new account is provisioned here.

Applied and tested.

Part of https://github.com/wellcometrust/wellcome-dns-infra/pull/120

### DIFF
--- a/accounts/rolesets.tf
+++ b/accounts/rolesets.tf
@@ -66,7 +66,9 @@ module "super_dev_roleset" {
     module.s3_releases_scala_typesafe.role_arn,
     module.s3_releases_scala_fixtures.role_arn,
 
-    # Route 53 - drupalinfra
+    # Route 53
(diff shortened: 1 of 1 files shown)


accounts/rolesets.tf
```diff
@@ -66,7 +66,9 @@ module "super_dev_roleset" {
     module.s3_releases_scala_typesafe.role_arn,
     module.s3_releases_scala_fixtures.role_arn,
 
-    # Route 53 - drupalinfra
+    # Route 53
+    "arn:aws:iam::267269328833:role/wellcomecollection-assume_role_hosted_zone_update",
+    # DERECATED
```
jamieparkinson, line 71:
```suggestion
    # DEPRECATED
```

alexwlchan: A bit more context here – even a link back to the private Slack thread – would make this comment more useful.

PR #27, opened 2020-06-30
Add VPC for iiif-services to be worked in by Digitati

kenoir:
For https://github.com/wellcomecollection/platform/issues/4605

### DIFF
--- a/critical/back_end/secrets.tf
+++ b/critical/back_end/secrets.tf
@@ -36,3 +36,13 @@ module "experience_logging_secrets" {
 
   secrets = local.logging_secrets
 }
+
+module "digirati_logging_secrets" {
+  source = "./modules/secrets/distributed"
+
+  providers = {
+    aws = aws.digirati
+  }
+
+  secrets = local.logging_secrets
+}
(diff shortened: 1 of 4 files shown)


critical/back_end/secrets.tf
```diff
@@ -34,5 +34,15 @@ module "experience_logging_secrets" {
     aws = aws.experience
   }
 
+  secrets = local.logging_secrets
+}
+
+module "digirati_logging_secrets" {
```
kenoir, line 40: Also adds logging secrets

critical/back_end/vpcs.tf
```diff
@@ -23,9 +23,13 @@ locals {
   experience_cidr_block_public  = cidrsubnet(local.experience_cidr_block_vpc, 1, 0)
   experience_cidr_block_private = cidrsubnet(local.experience_cidr_block_vpc, 1, 1)
 
-  developer_cidr_block_vpc     = "172.42.0.0/16"
+  developer_cidr_block_vpc     = "172.56.0.0/16"
```
donaldgray, line 26: This is the same as the digirati block - is that expected?

kenoir: Nope! Good catch - I actually fixed this before applying but forgot to commit that change - will push that update into master after this.

PR #31, opened 2020-07-13
CI for each repo should assume role

kenoir:
Instead of a fixed identity, CI should assume role allowing developers to mimic exactly the permissions and process of CI with regard to AWS permissions.


kenoir:
For https://github.com/wellcomecollection/platform/issues/4627

PR #32, opened 2020-07-13
Add release table permissions

kenoir:
Allow CI role to write to release table


kenoir:
For https://github.com/wellcomecollection/platform/issues/4627

PR #33, opened 2020-07-13
CI can get params

kenoir:
Ci role needs to be able to read SSM params.


kenoir:
For https://github.com/wellcomecollection/platform/issues/3806

PR #34, opened 2020-07-15
Allow workflow-support users to read from the storage-service

alexwlchan:
This will allow them to reingest items from the storage service through Goobi without downloading and reuploading the files.

Closes https://github.com/wellcomecollection/platform/issues/4665

### DIFF
--- a/accounts/iam_workflow_support.tf
+++ b/accounts/iam_workflow_support.tf
@@ -93,4 +93,33 @@ data "aws_iam_policy_document" "workflow_support" {
       "*",
     ]
   }
+
+  # Allow workflow-support users to browse the contents of the storage service,
+  # and download files in the digitised/ prefix.
+  #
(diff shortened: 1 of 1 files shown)


accounts/iam_workflow_support.tf
```diff
@@ -93,4 +93,33 @@ data "aws_iam_policy_document" "workflow_support" {
       "*",
     ]
   }
+
+  # Allow workflow-support users to browse the contents of the storage service,
+  # and download files in the digitised/ prefix.
+  #
+  # This is used to copy videos into the Goobi storage bucket for reingest.
+  # See https://wellcome.slack.com/archives/CN56BRQ5B/p1594804248033200
```
jamesgorrie, line 101: Worth summarising this in the ticket?

alexwlchan: I'm hoping there's a sufficient summary is in the comment; the link is just to tie it all together.

PR #39, opened 2020-07-23
Add CloudFront distros for IIIF APIs

kenoir:
For https://github.com/wellcomecollection/platform/issues/4603

When this is released the following Loris specific CloudFront distro can be removed (along with the cert it uses, and that cert can then be definitively referenced in terraform here).

https://github.com/wellcomecollection/loris-infrastructure/blob/master/terraform/cloudfront.tf

### DIFF
--- a/cloudfront/iiif/main.tf
+++ b/cloudfront/iiif/main.tf
@@ -1,7 +1,7 @@
 locals {
   subdomain_modifier = var.environment == "prod" ? "" : "-${var.environment}"
 
-  distro_alias = "iiif${local.subdomain_modifier}.wellcomecollection.org"
+  distro_alias = "iiif-${var.environment}.wellcomecollection.org"
 
   dds_domain   = "dds${local.subdomain_modifier}.dlcs.io"
   iiif_domain  = "iiif${local.subdomain_modifier}.dlcs.io"
--- a/cloudfront/iiif/main.tf
+++ b/cloudfront/iiif/main.tf
@@ -1,0 +1,439 @@
+locals {
+  subdomain_modifier = var.environment == "prod" ? "" : "-${var.environment}"
+
+  distro_alias = "iiif${local.subdomain_modifier}.wellcomecollection.org"
+
+  dds_domain   = "dds${local.subdomain_modifier}.dlcs.io"
+  iiif_domain  = "iiif${local.subdomain_modifier}.dlcs.io"
+  loris_domain = "iiif-origin.wellcomecollection.org"
+  dlcs_domain  = "dlcs.io"
+}
+
+resource "aws_cloudfront_distribution" "iiif" {
+  aliases = [
+    local.distro_alias
+  ]
+
+  origin {
+    domain_name = local.dds_domain
+    origin_id   = "dds"
+
+    custom_origin_config {
+      origin_protocol_policy = "match-viewer"
+      origin_ssl_protocols = ["TLSv1"]
+
+      http_port = 80
+      https_port = 443
+    }
+  }
+
+  origin {
+    domain_name = local.dlcs_domain
+    origin_id   = "dlcs"
+
+    custom_origin_config {
+      origin_protocol_policy = "match-viewer"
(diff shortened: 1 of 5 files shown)


cloudfront/iiif/main.tf
```diff
@@ -0,0 +1,439 @@
...
+    origin_id   = "iiif"
+
+    custom_origin_config {
+      origin_protocol_policy = "match-viewer"
+      origin_ssl_protocols = ["TLSv1"]
+
+      http_port = 80
+      https_port = 443
+    }
+  }
```
alexwlchan, line 67: It might be overkill here, but it occurs to me that this would be a good use for dynamic blocks in Terraform.

cloudfront/iiif/main.tf
```diff
@@ -0,0 +1,439 @@
...
+        forward = "all"
+      }
+    }
+
+    min_ttl                = 0
+    default_ttl            = 86400
+    max_ttl                = 31536000
+
+    viewer_protocol_policy = "redirect-to-https"
+  }
```
alexwlchan, line 267: This definitely feels like a place for dynamic blocks, if we can make them work.

cloudfront/main.tf
```diff
@@ -0,0 +1,24 @@
...
+
+module "iiif-test" {
+  source = "./iiif"
+
+  environment = "test"
+  acm_certificate_arn = local.acm_cert_arn
+}
+
+locals {
+  acm_cert_arn = "arn:aws:acm:us-east-1:760097843905:certificate/04bb4447-b501-453a-804e-411d3f660a74"
```
alexwlchan, line 23: FWIW, you can also look up ACM certificates by domain name: https://www.terraform.io/docs/providers/aws/d/acm_certificate.html

jamieparkinson:
Can you reduce the duplication by using a dynamic block `for_each`ing over the `path_patterns`?
Can you get rid of this? https://github.com/wellcomecollection/loris-infrastructure/blob/master/terraform/cloudfront.tf

cloudfront/provider.tf
```diff
@@ -0,0 +1,8 @@
+provider "aws" {
+  region  = "eu-west-1"
```
alexwlchan, line 2: I thought CloudFront distros had to live in us-east-1?

cloudfront/terraform.tf
```diff
@@ -0,0 +1,12 @@
+terraform {
+  required_version = ">= 0.9"
+
+  backend "s3" {
+    bucket         = "wellcomecollection-platform-infra"
+    key            = "terraform/platform-infrastructure/cloudfront.tfstate"
```
alexwlchan, line 6: 👌

alexwlchan:
Ah, I guess there's no way to nest `custom_origin_config`?

alexwlchan:
I managed to get something “working”: https://github.com/wellcomecollection/platform-infrastructure/pull/40

Or at least, it says there’s nothing to change when I run a plan – does the problem happen when you try to apply?

cloudfront/main.tf
kenoir: This is awkward if there are duplicated for the same name.

cloudfront/provider.tf
kenoir: They are "global" - certs for them must live in `us-east-1`.

kenoir:
> Or at least, it says there’s nothing to change when I run a plan – does the problem happen when you try to apply?

The issue is on creation as described in https://github.com/terraform-providers/terraform-provider-aws/issues/10088

kenoir:
> I assume that Cache-Control public headers from the origin will always override default_ttl in Cloudfront?

I believe so.

kenoir:
> Can you get rid of this? https://github.com/wellcomecollection/loris-infrastructure/blob/master/terraform/cloudfront.tf

Yes, when we've moved to the new distro.

kenoir:
It will take a few hours to ensure that the new distro is stable before we think about disabling and removing the old one.

PR #46, opened 2020-08-13
Use our own NAT instances in the storage VPC

alexwlchan:
Closes https://github.com/wellcomecollection/platform/issues/4666

This runs an EC2 instance in the storage VPC that acts as a NAT instance, so we don't have to run a NAT Gateway. This will save us the eye-watering cost of sending everything that goes to Azure through the managed Gateway ($0.048/GB sent).

Notes:

* I haven't turned this on for every VPC, because that seems somewhat disruptive. It took several attempts to get working on the first try; we should wait until we have time to debug any issues that arise. We can opt a VPC into using this by setting `use_nat_instances = true` (default false).

* Unlike the managed NAT Gateway, the bandwidth on NAT instances depends on the bandwidth of the underlying instances. I picked a t3.medium ($0.0456/hr) because it’s about the same as a managed NAT Gateway ($0.048/hr), but without the per-GB costs. We can review that if necessary, e.g. if the instance size becomes a bottleneck to the initial Azure migration.

* To use the `count` on modules, I had to upgrade to Terraform 0.13.0. I had to do some manual cleanups of the state file (see https://github.com/hashicorp/terraform/issues/25752), but it seems to be working.


alexwlchan:
Possibly useful for reviewers is Amazon’s guide to NAT instances: https://docs.aws.amazon.com/vpc/latest/userguide/VPC_NAT_Instance.html

README.md
```diff
@@ -20,3 +20,8 @@ Wellcome Collection common infrastructure.
 
 - **digitisation_infra**: Digital production infrastructure.
 
+
+
+## Third-party code
+
+*   The `nat_instance` module used in `critical/back_end` is based on [int128/terraform-aws-nat-instance](https://github.com/int128/terraform-aws-nat-instance) at commit 1a6398a, used under the Apache licence.
```
jamieparkinson, line 27:
```suggestion
*   The `nat_instance` module used in `critical/back_end` is based on [int128/terraform-aws-nat-instance](https://github.com/int128/terraform-aws-nat-instance/tree/1a6398adbe242fc8d3ca1efe22090ad7fe195a98), used under the Apache licence.
```

critical/back_end/modules/nat_instance/main.tf
```diff
@@ -0,0 +1,186 @@
...
+}
+
+resource "aws_route" "nat_instance" {
+  route_table_id         = var.private_subnet_route_table_id
+  destination_cidr_block = "0.0.0.0/0"
+  network_interface_id   = aws_network_interface.nat_instance.id
+}
+
+# AMI of the latest Amazon Linux 2
+data "aws_ami" "nat_instance" {
```
jamieparkinson, line 50: How come we manually set up the NAT stuff on an AL2 AMI rather than using one of Amazon's NAT AMIs?

critical/back_end/modules/nat_instance/main.tf
```diff
@@ -0,0 +1,186 @@
...
+
+resource "aws_launch_template" "nat_instance" {
+  name_prefix = var.name
+  image_id    = data.aws_ami.nat_instance.id
+
+  iam_instance_profile {
+    arn = aws_iam_instance_profile.nat_instance.arn
+  }
+
+  instance_type = "t3.medium"
```
jamieparkinson, line 83: Maybe a variable?

critical/back_end/modules/nat_instance/main.tf
alexwlchan: That’s the way it was done in the module I copied, and I didn't want to change it more than necessary.

PR #56, opened 2020-09-28
Break up the 'accounts' monolith configuration into per-account configurations

alexwlchan:
Spun out of https://github.com/wellcomecollection/platform/issues/4802

Currently the managed roles and IAM policies for every account sit in one Terraform configuration. This is a dangerous abstraction: I shouldn't run the risk of breaking something in *platform* if I want to change something in *storage*, and so on. It would be better if each account was managed in a separate configuration.

This PR illustrates what that might look like, by splitting out the digirati account into a new configuration. I haven't removed it from the main configuration yet, so we can drop this change non-destructively.

Before I start breaking things or doing more work, I'd like to get other opinions: **is this a good idea? Should I do it for the remaining accounts?**

Assuming it it, I'll finish this off, then move the VPC and CloudHealth definitions to live inside these configurations. Each configuration would contain the "base" state for each of our AWS accounts.

—

I've also written a Python script that works out what import statements to create based on the existing state in the `accounts` configuration

<details>
```python
#!/usr/bin/env python

import json
import shlex
import sys
from pprint import *

name = sys.argv[1]
role_arn = sys.argv[2]

tfstate = json.load(open('/Users/chana/Desktop/accounts.tfstate.json'))

for role_name in (
    "admin",
    "billing",
    "ci",
    "developer",
    "monitoring",
    "publisher",
    "read_only",
):
    terraform_id = shlex.quote(f"{name}-{role_name}")
    print(
        f"terraform import module.{name}_account.module.{role_name}_role.aws_iam_role.role {terraform_id}"
    )

for role_name, policy_id in [
    ('admin', 'admin'),
    ('billing', 'admin'),
    ('ci', 'ci_permissions'),
    ("developer", "developer_go_hog_wild"),
    ("developer", "developer_iam_user_group_deny"),
    ("developer", "developer_restrict_destructive_actions"),
    ('monitoring', 'billing'),
    ('monitoring', 'ecs'),
    ('publisher', 'ecr_power_user'),
    ('publisher', 'ssm_parameter_store_power_user'),
    ('read_only', 'test-attach'),
]:
    matching_resources = [
        r
        for r in tfstate['resources']
        if r.get('module') == f'module.{name}_account.module.{role_name}_role_policy' and r['type'] in {'aws_iam_role_policy', 'aws_iam_role_policy_attachment'} and r.get('name') == policy_id
    ]
    
    if len(matching_resources) == 1 and len(matching_resources[0]['instances']) == 1:
        resource_id = matching_resources[0]['instances'][0]['attributes']['id']
    else:    
        pprint(matching_resources)
        assert 0
    
    terraform_id = shlex.quote(resource_id)
    if role_name == 'read_only':        
        print(
            f"terraform import module.{name}_account.module.{role_name}_role_policy.aws_iam_role_policy_attachment.{policy_id} {name}-{role_name}/arn:aws:iam::aws:policy/ReadOnlyAccess"
        )
    else:
        print(
            f"terraform import module.{name}_account.module.{role_name}_role_policy.aws_iam_role_policy.{policy_id} {terraform_id}"
        )
```
</details>


kenoir:
> Assuming it it, I'll finish this off, then move the VPC and CloudHealth definitions to live inside these configurations. Each configuration would contain the "base" state for each of our AWS accounts.

This is part of the breaking up critical work then? Yes - in which case it makes sense to do this alongside that change.

alexwlchan:
> This is part of the breaking up critical work then?

Yup, so the “long term” plan (later today) is

```
accounts/
  catalogue/
    account.tf
    vpc.tf
    cloudhealth.tf
  digirati/
    account.tf
    vpc.tf
    cloudhealth.tf
  ...
```

alexwlchan:
This is now ready for review – the "accounts" monolith has been broken up into per-account subfolders, each of which has a smaller structure. What used to be "accounts" is now "platform".

alexwlchan:
It's probably easier to review if you just browse the new hierarchy: https://github.com/wellcomecollection/platform-infrastructure/tree/split-out-accounts/accounts

kenoir (approved):
Assuming this applies with a no-op!

alexwlchan:
Yup, already sorted!

PR #58, opened 2020-09-28
Define VPCs in different accounts in different configurations

alexwlchan:
Part of https://github.com/wellcomecollection/platform/issues/4802. Already applied as a no-op change.

This means that changing one VPC won't break VPCs in a different account, e.g. tweaking the *ci* VPC won't break the *storage* VPC. The VPCs are now defined in `accounts/<name>/vpcs.tf`.

All the VPC IDs/subnets are re-exported through the *critical/back_end* module, so this won't break any existing code – but I will go around and update other repos to point at the per-account config once this is merged.


kenoir (approved):
Terraform change should be a no-op obvs.

alexwlchan:
Yup, already done.

PR #64, opened 2020-10-06
Add a Buildkite agent hook that cleans up any leftover containers

alexwlchan:
tl;dr: all our Buildkite agents now run a script that stops all the running containers before they run a job, so any state leftover from previous jobs gets cleared out. This means instances don't have to be restarted after every job.

You can see the first example here: this agent didn't have to redownload Docker images and Scala libraries before running the autoformat task. https://buildkite.com/organizations/wellcomecollection/agents/8fcdc05c-ef00-4164-bd6c-e725f1836322

The comments in the code explain the exact mechanism.

A tentative first step towards https://github.com/wellcomecollection/platform/issues/4818


builds/buildkite_agent_hook.sh
```diff
@@ -0,0 +1,17 @@
...
+set -o errexit
+set -o nounset
+
+echo "Hello! I am the agent hook downloaded from S3!"
+
+echo "These are the containers which are running, probably from a previous job:"
+docker ps
+
+echo "I'm going to stop these containers now, so this job has a clean start."
+docker stop $(docker ps -q) || true
```
jamieparkinson, line 15: Given that we are not bothered about stopping containers gracefully, could this be `docker kill`?

jamieparkinson (approved):

builds/buildkite_agent_hook.sh
alexwlchan: Sounds good. I hadn't heard of the command before; will change.

PR #66, opened 2020-10-06
Add autoformatting in Buildkite

alexwlchan:
This config is copied from the catalogue repo.

Closes https://github.com/wellcomecollection/platform/issues/4808

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,0 +1,3 @@
+steps:
+  - command: .buildkite/scripts/run_autoformat.py
+    label: "autoformat"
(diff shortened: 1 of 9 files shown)


kenoir (approved):
This is good as is.

Just to note that it may be possible to simplify the build here considerably by pushing the format step inside a docker container (use one of the stock terraform images from hashicorp and run `terraform fmt`), then committing using this plugin: https://github.com/thedyrt/git-commit-buildkite-plugin

At your discretion you might want to try it.

alexwlchan:
I will leave it as-is for now, so builds in this repo start working again, but will look at that if I have time.

PR #76, opened 2020-11-17
Mirror the Docker Hub images we care about to ECR

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/4879

Hopefully the config at the top of the script makes it pretty self-explanatory how to add new images. Given all our CI runs in the platform account, we might not have to worry about pushing images into multiple accounts, but that's on the table if we ever want it.


kenoir (approved):
This looks good - quite an aside but I wonder if it would also have been possible to do this only using terraform with the docker hub provider?

edit: nope, the ECR resources in the AWS provider don't include images.

PR #78, opened 2020-11-19
Ensure Buildkite can assume identity CI role

kenoir:
Needed for builds on https://github.com/wellcomecollection/identity


builds/terraform/buildkite.tf
```diff
@@ -85,6 +85,7 @@ data "aws_iam_policy_document" "ci_permissions" {
       local.account_ci_role_arn_map["storage"],
       local.account_ci_role_arn_map["experience"],
       local.account_ci_role_arn_map["workflow"],
+      local.account_ci_role_arn_map["identity"],
```
alexwlchan, line 88: Minor: you might be able to use some TF collection functions to reduce repetition here, e.g.

```
concat(
  [local.platform_read_only_role_arn],
  values(local.account_ci_role_arn_map)
)
```

but no biggie.

PR #82, opened 2020-11-25
Adds SES verification and DKIM records for Identity project

kenoir:
Verifies SES as capable of using `wellcomecollection.org` to send e-mail. 

See https://en.wikipedia.org/wiki/DomainKeys_Identified_Mail to understand what's going on with the DKIM records!

### DIFF
--- a/dns/locals.tf
+++ b/dns/locals.tf
@@ -7,7 +7,7 @@ locals {
   // want to allow the outputs from that stack to set those details here in order
   // to more tightly control of wellcomecollection.org records in this stack.
   identity_ses_txt_records = data.terraform_remote_state.identity.outputs.wellcomecollection_org_ses_vertification_token[["records"]]
-  identity_dkim_cname = data.terraform_remote_state.identity.outputs.wellcomecollection_org_ses_dkim_tokens
+  identity_dkim_cname      = data.terraform_remote_state.identity.outputs.wellcomecollection_org_ses_dkim_tokens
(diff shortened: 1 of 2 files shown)


kenoir:
This PR is deliberately missing some locals - awaiting these being available in the identity static terraform state.

dns/locals.tf
```diff
@@ -1,4 +1,14 @@
 
 locals {
   identity_zone_name_servers = data.terraform_remote_state.identity.outputs.identity_zone_name_servers
+
+  // The outputs from the identity stack contain enough information to construct
+  // the whole record (including type, and name) which is useful _but_ we don't
+  // want to allow the outputs from that stack to set those details here in order
+  // to more tightly control of wellcomecollection.org records in this stack.
+  identity_ses_txt_records = data.terraform_remote_state.identity.outputs.wellcomecollection_org_ses_vertification_token[["records"]]
```
alexwlchan, line 9: Possibly daft question: why are there two share brackets around records?

alexwlchan (approved):
Looks good. We do sometimes put the DNS verification alongside the thing it's verifying (e.g. https://github.com/wellcomecollection/archivematica-infrastructure/blob/development/terraform/infra/certs.tf) but I think having it here is fine.

dns/locals.tf
kenoir: tyop

kenoir: works though ... 🤷

PR #84, opened 2020-11-27
Refining permissions for API Gateway and Lambda Function deployments for Identity

danielgrant:
The deployment process via Buildkite requires the ability to modify AWS Lambda Function's and API Gateway resources. The permissions for this are already in place, but the ARN targets are invalid. This PR uses the correct ARN's.


jamesgorrie (approved):
Any chance you could give a little more context as to why we need this change in the description?

danielgrant:
> Any chance you could give a little more context as to why we need this change in the description?

Have done.

kenoir:
This is applied.

PR #85, opened 2020-11-27
Rescoping API Gateway permissions for Identity CI user

danielgrant:
The current ARN for the API Gateway IAM permissions for the Identity CI user target the `/stages` resources only. This is incorrect, wider scope is required. This change widens that scope to all API Gateway resources.


kenoir:
This has been applied.

kenoir (approved):

PR #86, opened 2020-11-27
Adding Lambda 'GetAlias' to Identity CI

danielgrant:
In order to complete the Buildkite deployment pipeline, the Identity CI user must be able to determine if a Lambda alias already exists.


jamesgorrie:
Applied and working.

PR #98, opened 2021-01-25
Provision mirrored images

kenoir:
Provision the images we mirror from Docker hub in Terraform, update the mirroring script accordingly.

Note: This terraform is applied.


alexwlchan (approved):

kenoir:
Yes it may be possible to sync the list of images from terraform outputs into the python script - I didn't feel like it was worth it (yet).

PR #101, opened 2021-02-04
Improve edge-lambda build

kenoir:
In order to add the appropriate path rewrites routes under `iiif.wellcomecollection.org`, we need to make use of lambda@edge. 

For type safety and similarity to other current work we should to compile these lambdas from typescript (see the [wellcomecollection.org repository](https://github.com/wellcomecollection/wellcomecollection.org/tree/master/cache/edge_lambdas)).

This PR replaces existing `.js` lambdas with `.ts` versions and adds a build and deploy mechanism in CI (previously this had to be done by a developer).


cloudfront/lambdas/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "dockerTestLocal": "yarn dockerBuildLocal && docker run weco_cf_lambdas yarn test",
+    "dockerDeployLocal": "yarn dockerBuildLocal && docker run weco_cf_lambdas -v ~/.aws:/root/.aws yarn deploy"
+  },
+  "devDependencies": {
+    "@types/aws-lambda": "^8.10.71",
+    "@types/jest": "^26.0.20",
+    "@types/node": "^14.14.25",
+    "jest": "^26.6.3",
+    "ts-jest": "^26.5.0",
+    "tsc": "^1.20150623.0",
```
jamesgorrie, line 24: I assume this is a mistake

cloudfront/lambdas/tsconfig.build.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "noEmit": false
+  },
+  "exclude": ["node_modules","**/*.test.ts", "**/test_*.ts"]
```
jamieparkinson, line 6: fwiw, credit where it's due, if you've copied this idea of a `tsconfig.build.json` extending a base config from the example in weco.org - I copied it off digirati!

jamesgorrie (approved):
I'd say good to go in like this

cloudfront/lambdas/package.json
kenoir: Which what?

jamieparkinson: tsc dependency shouldn't be there (tsc is part of typescript) https://www.npmjs.com/package/tsc

accounts/platform/.terraform.lock.hcl
```diff
@@ -1,38 +0,0 @@
-# This file is maintained automatically by "terraform init".
```
alexwlchan, line 1: Why have you deleted this?

accounts/platform/iam_platform_ci.tf
```diff
@@ -0,0 +1,23 @@
+resource "aws_iam_role_policy" "platform_ci" {
+  role   = module.aws_account.ci_role_name
+  policy = data.aws_iam_policy_document.platform_ci.json
+}
+
+data "aws_iam_policy_document" "platform_ci" {
+  dynamic "statement" {
+    for_each = [
+      "arn:aws:s3:::wellcomecollection-edge-lambdas"
```
alexwlchan, line 9: I assume these are being uploaded to a different bucket to our usual Lambdas because they need to live in us-east-1 with CloudFront?

cloudfront/iiif/edge-lambda/src/dlcs_path_rewrite.test.ts
```diff
@@ -0,0 +1,37 @@
...
+    },
+    {
+      in: '/image/B0009852.JPG/full/1338%2C/0/default.jpg',
+      out: '/B0009852/full/1338%2C/0/default.jpg'
+    },
+    {
+      in: '/image/B0009853/full/1338%2C/0/default.jpg',
+      out: '/B0009853/full/1338%2C/0/default.jpg'
+    }
+  ]
```
alexwlchan, line 24: Having not looked at the implementation yet, I'd suggest a couple of other test cases:

* What if the output format is something other than `.jpg`, e.g. `/default.png`
* What if the region isn't `full`? e.g. `B0009853/400,400,200,200`
* What if the quality isn't `default`? e.g. `/gray.jpg`

cloudfront/iiif/edge-lambda/src/dlcs_path_rewrite.ts
```diff
@@ -0,0 +1,11 @@
+import {
+    CloudFrontRequestHandler,
+} from 'aws-lambda';
+
+export const request: CloudFrontRequestHandler = (event, context, callback) => {
+    const request = event.Records[0].cf.request;
+
+    request.uri = request.uri.replace(/^\/image\/([A-Za-z0-9]+)(?:\.[Jj][Pp][Gg])?(\/.*)/gm, '/$1$2')
```
alexwlchan, line 8: Can you provide a comment explaining how this regex works? What bit of the original URL is it meant to capture and replace? My initial read of it was wrong, I think – I thought the `[Jj][Pp][Gg]` was on the end of `default.jpg`, not the image ID.

cloudfront/iiif/edge-lambda/src/test_event_request.ts
```diff
@@ -0,0 +1,28 @@
...
+        request: {
+          uri,
+          querystring: querystring || '',
+          method: 'GET',
+          clientIp: '2001:cdba::3257:9652',
+          headers: {},
+        },
+      },
+    },
+  ],
```
alexwlchan, line 25: I assume this is from https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/lambda-event-structure.html#example-origin-request; having a link in the code to the event this is based on would be helpful.

kenoir:
For https://github.com/wellcomecollection/platform/issues/5007

accounts/platform/iam_platform_ci.tf
kenoir: Yes.

accounts/platform/.terraform.lock.hcl
kenoir: It should not be deleted - I needed to remove it temporarily to get an update that fixed an issue with the vpc endpoint module.

kenoir: Have re-added. Thanks.

cloudfront/iiif/edge-lambda/src/dlcs_path_rewrite.ts
kenoir: Good idea. This is on the verge of "regex too complicated, simple thing instead".

cloudfront/iiif/edge-lambda/src/test_event_request.ts
kenoir: It's cargo-culted from the wellcomecollection.org repo - though it looks to have been sourced from there originally. Will add comment.

cloudfront/iiif/edge-lambda/src/dlcs_path_rewrite.ts
alexwlchan: Maybe something like:

```
function convertLorisPathToDlcsPath(path: string) {
    const pathParts: Array<string> = path.split(",", 2)

    if (pathParts.length === 3 || pathParts[0] === "image") {
        const lorisImageId = pathParts[1]
        const dlcsImageId = lorisImageId.replace(/\.[Jj][Pp][Gg]/, "")
        path.replace(lorisImageId, dlcsImageId)
    } else {
        path
    }
} 
```

cloudfront/iiif/edge-lambda/src/dlcs_path_rewrite.test.ts
kenoir: Good point. Especially as i'm going to add to the things that get rewritten soon!

PR #102, opened 2021-02-05
Split our CloudFront distributions into a different stack per subdomain

alexwlchan:
This limits the potential damage from any Terraform change to one subdomain – rather than modifying all CloudFront distros at once, we only modify a subdomain at a time.

This is what the top level of the "cloudfront" directory now looks like:

Screenshot 2021-02-05 at 18 47 55

These changes will all plan/apply with no changes.

Closes https://github.com/wellcomecollection/platform/issues/5008


kenoir (approved):
Brilliant 👍 

Only comment is i'm not a fan of the wc.org contraction (as it's not actually our domain, and does resolve to someoneelse'ss site) outside of typing it in Slack. Shouldn't block merging though.

PR #106, opened 2021-02-08
Improve usability of iiif CloudFront distribution

kenoir:
The current CloudFront distribution(s) that server iiif.wellcomecollection.org over the environments can be difficult to understand and to modify (especially with regard to modifying environments separately). 

As part of moving from Loris to DLCS and as part of the upcoming redirection work as we decommission wellcomelibrary.org, it's useful to improve the clarity and ease of use of these distributions.

Part of https://github.com/wellcomecollection/platform/issues/5007


cloudfront/iiif.wellcomecollection.org/terraform/cf_behaviours.tf
```diff
@@ -0,0 +1,44 @@
...
+  wellcome_images_path_patterns = [
+    "image/V00*",
+    "image/L00*",
+    "image/M00*",
+    "image/B00*",
+    "image/N00*",
+    "image/A00*",
+    "image/W00*",
+    "image/S00*"
+  ]
```
alexwlchan, line 11: Minor: could we sort these alphabetically? The order looks arbitrary; alphabetical is slightly more readable.

```suggestion
  wellcome_images_path_patterns = [
    "image/A00*",
    "image/B00*",
    "image/L00*",
    "image/M00*",
    "image/N00*",
    "image/S00*"
    "image/V00*",
    "image/W00*",
  ]
```

It also makes it easier to compare these to a list of prefixes from Miro – e.g. AS or FP. Do we need to include those as well?

cloudfront/iiif.wellcomecollection.org/terraform/cloudfront_distro/main.tf
```diff
@@ -122,244 +53,37 @@ resource "aws_cloudfront_distribution" "iiif" {
...
+  dynamic "ordered_cache_behavior" {
+    for_each = var.behaviours
+    content {
+      path_pattern     = ordered_cache_behavior.value["path_pattern"]
+      allowed_methods  = ["DELETE", "GET", "HEAD", "OPTIONS", "PATCH", "POST", "PUT"]
+      cached_methods   = ["GET", "HEAD", "OPTIONS"]
+      target_origin_id = ordered_cache_behavior.value["target_origin_id"]
+
+      forwarded_values {
+        query_string = true
```
jamesgorrie, line 65: Do we use the query string?

cloudfront/iiif.wellcomecollection.org/terraform/cf_behaviours.tf
```diff
@@ -0,0 +1,44 @@
...
+      headers          = []
+      cookies          = "none"
+      lambdas = [
+        {
+          event_type = "origin-request"
+          lambda_arn = local.dlcs_path_rewrite_arn_latest
+        }
+      ]
+    }
+  ]
```
alexwlchan, line 38: 👍 

These `for` loops are really nice.

jamesgorrie (approved):
tidy!

cloudfront/iiif.wellcomecollection.org/terraform/cloudfront_distro/main.tf
```diff
@@ -1,100 +1,31 @@
...
-        "TLSv1.1",
-        "TLSv1.2",
-      ]
-
-      http_port  = 80
-      https_port = 443
+  dynamic "origin" {
+    for_each = var.origins
+    content {
+      origin_id   = origin.value["origin_name"]
```
alexwlchan, line 14: minor: could we call this key `"origin_id"` for consistency?

cloudfront/iiif.wellcomecollection.org/terraform/cloudfront_distro/main.tf
```diff
@@ -122,244 +53,37 @@ resource "aws_cloudfront_distribution" "iiif" {
...
-
-    min_ttl     = 604800
-    default_ttl = 86400
-    max_ttl     = 31536000
 
-    viewer_protocol_policy = "redirect-to-https"
-  }
+      min_ttl     = 604800
+      default_ttl = 86400
+      max_ttl     = 31536000
```
alexwlchan, line 83: minor: I know you haven't changed these, but while you're here could we update these to make their values more meaningful?

```suggestion
      min_ttl     = 7 * 24 * 60 * 60
      default_ttl = 24 * 60 * 60
      max_ttl     = 365 * 24 * 60 * 60
```

Possibly even pull `24 * 60 * 60` into a `days` local.

cloudfront/iiif.wellcomecollection.org/terraform/main.tf
```diff
@@ -11,21 +14,17 @@ module "iiif-stage" {
...
 }
 
 module "iiif-test" {
   source = "./cloudfront_distro"
 
   environment         = "test"
   acm_certificate_arn = module.cert.arn
+
+  origins    = local.origins["test"]
+  behaviours = local.behaviours
```
alexwlchan, line 29: Minor: should this be a local called `test_behaviours`, and defined where the other two variables are? That way all three are defined in the same place, and it's clearer there if you want to change the behaviour of the test distribution.

cloudfront/iiif.wellcomecollection.org/terraform/outputs.tf
```diff
@@ -0,0 +1,7 @@
+# This output is used to inform the developer making changes which is the latest version
+# The latest version will be used in the stage env by default, but should be set
+# manually in prod, after verifying correct behaviour.
```
alexwlchan, line 3: Good comment. 👍

cloudfront/iiif.wellcomecollection.org/terraform/cf_behaviours.tf
kenoir: > Do we need to include those as well?
There don't seem to be any images that require registration in the AS/FP prefix range, so they wouldn't strictly be required here.

cloudfront/iiif.wellcomecollection.org/terraform/cloudfront_distro/main.tf
kenoir: but then all the text doesn't line up nicely :(

👍 will fix this

cloudfront/iiif.wellcomecollection.org/terraform/cloudfront_distro/main.tf
kenoir: I don't actually think we do, but this is the same as currently configured, so i'm reluctant to fix it in this PR.

PR #107, opened 2021-02-10
Rejig DNS / Cloudfront directories

jamesgorrie:
To be able to test if we can hijack the `HOST` header and start redirecting traffic.

Waiting on D&T to allow permissions to `RequestCertificate`


dns/modules/ssl_cert/README.md
```diff
@@ -0,0 +1,13 @@
+# ssl_cert
```
kenoir, line 1: There is already a certificate module in this repo I can see ... I wonder whether we can just adapt that one, rather than bringing in this one ... i'll have a go and push that in another draft PR to have a look at.

kenoir: Like this: https://github.com/wellcomecollection/platform-infrastructure/pull/111

kenoir (approved):

kenoir:
Part of https://github.com/wellcomecollection/platform/issues/4977

PR #108, opened 2021-02-17
Add some lifecycle rules to the platform-infra bucket

alexwlchan:
We shouldn't be keeping everything that goes in this bucket indefinitely. This is particularly driven by https://github.com/wellcomecollection/platform/issues/5025, which will start adding 1GB+ snapshots of the Sierra data to this bucket, and we don't need to keep multiple copies of that.

Because this is a potentially destructive change, I'm not going to apply it until it's reviewed.

### DIFF
--- a/critical/back_end/s3_platform_infra.tf
+++ b/critical/back_end/s3_platform_infra.tf
@@ -21,16 +21,6 @@ resource "aws_s3_bucket" "platform_infra" {
       storage_class = "STANDARD_IA"
     }
 
-    noncurrent_version_transition {
-      days          = 30
-      storage_class = "STANDARD_IA"
-    }
-
-    noncurrent_version_transition {
-      days          = 60
-      storage_class = "GLACIER"
-    }
-
     noncurrent_version_expiration {
       days = 90
     }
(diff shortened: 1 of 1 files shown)


critical/back_end/s3_platform_infra.tf
```diff
@@ -12,6 +12,30 @@ resource "aws_s3_bucket" "platform_infra" {
...
+
+    noncurrent_version_transition {
+      days          = 60
+      storage_class = "GLACIER"
+    }
+
+    noncurrent_version_expiration {
+      days = 90
+    }
+  }
```
jamesgorrie, line 37: What benefits do we get by going through this lifecycle over just expiring the versions?

alexwlchan: Slightly lower costs, although tbh it's just copy/pasta from whatever example I got this from. I can remove it.

jamesgorrie: 🍝

jamesgorrie (approved):

jamesgorrie:
A thought is that there are some terraform states that don't get updated for ages (platform-infrastructure). I don't think I can find a reason why this is a problem, but it just crossed my mind.

alexwlchan:
> A thought is that there are some terraform states that don't get updated for ages (platform-infrastructure). I don't think I can find a reason why this is a problem, but it just crossed my mind.

The useful time to roll back to a previous version of state is straight after something went wrong; I think it's very unlikely we'd want to roll back to state from more than three months ago.

PR #110, opened 2021-02-23
Add wellcomelibrary.org CloudFront distribution

kenoir:
Adds a skeleton CloudFront distribution for wellcomelibrary.org.

Part of https://github.com/wellcomecollection/platform/issues/4977

**Note:** This change has been applied.

### DIFF
--- a/cloudfront/wellcomelibrary.org/terraform/cert.tf
+++ b/cloudfront/wellcomelibrary.org/terraform/cert.tf
@@ -9,6 +9,8 @@ module "cert" {
     aws     = aws.us_east_1
     aws.dns = aws.dns
   }
+
+  subject_alternative_names = ["*.wellcomelibrary.org"]
 }
 
 data "aws_route53_zone" "zone" {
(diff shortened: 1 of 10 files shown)


cloudfront/wellcomelibrary.org/terraform/cert.tf
```diff
@@ -0,0 +1,20 @@
...
+  domain_name = "wellcomelibrary.org"
+
+  zone_id = data.aws_route53_zone.zone.id
+
+  providers = {
+    aws     = aws.us_east_1
+    aws.dns = aws.dns
+  }
+
+  subject_alternative_names = ["*.wellcomelibrary.org"]
```
jamesgorrie, line 13: Do we want to do this and not have a cert per subdomain? I keep toing and froing about having a cert for each subdomain, which does feel less couple-y - and think I prefer having a cert per sub-domain.

jamesgorrie (approved):

cloudfront/wellcomelibrary.org/terraform/cert.tf
kenoir: Fewer certs makes sense I think, less things that require renewing. I'm not sure. We can always add more certs later.

PR #116, opened 2021-03-01
Redirect Wellcome Library Item pages 

kenoir:
Identify requests for item pages, perform a catalogue API lookup and path rewrite.

Part of https://github.com/wellcomecollection/platform/issues/5059


cloudfront/wellcomelibrary.org/edge-lambda/src/wellcomeLibraryRedirect.ts
```diff
@@ -1,20 +1,58 @@
...
+}
+
+// TODO: Account for multiple results by looking for matching sierra entry
+// TODO: Account for situations where results overflow page
+async function getWork(bNumber: string) {
+  const apiPath = `/works?identifiers=${bNumber}`
+  const results = await apiFetch(apiPath)
+
+  const resultList = results.data as CatalogueResultsList
+  assert(resultList.results.length >= 1, "No matching Catalogue API results found")
```
alexwlchan, line 20: Is an assertion the best behaviour here? Wouldn't it be better to return a 404 to the user?

cloudfront/wellcomelibrary.org/edge-lambda/src/wellcomeLibraryRedirect.ts
```diff
@@ -1,20 +1,58 @@
...
+
+  return resultList.results[0]
+}
+
+// TODO: Properly test for all these cases
+function getBnumberFromPath(path: string) {
+  const splitPath = path.split("/")
+  const bNumberRegexp = /^b[0-9]{8}/
```
alexwlchan, line 28: This is wrong – b numbers can end with an `x` as the check digit.

alexwlchan: Also, it would be handy to know what sort of path you expect this to match.

Possibly move the extraction logic into a separate file? It's just string parsing, could be managed/tested on its own.

cloudfront/wellcomelibrary.org/edge-lambda/src/wellcomeLibraryRedirect.ts
jamesgorrie: Assert exits with a non-0 code, which in these lambdas, we should never do as that will send the 500 on the cache layer.

cloudfront/wellcomelibrary.org/edge-lambda/src/wellcomeLibraryRedirect.ts
jamesgorrie: [Assuming that this is where we'll do it though](https://github.com/wellcomecollection/platform-infrastructure/pull/116/files#diff-cad288977e3edeafbbc2ad132d4fd147a6762ba2623c35f9fc6379e3d638d2d3R38)

kenoir: >This is wrong – b numbers can end with an x as the check digit.

I'm not sure if they can in this context, this is redirecting a specific pattern from the wellcomelibrary site. I'm not sure that there would be any links of this type pointing at un-check-digited bnumbers (I couldn't find any when I went a-googling).

cloudfront/wellcomelibrary.org/edge-lambda/src/wellcomeLibraryRedirect.ts
kenoir: Obvs.

cloudfront/wellcomelibrary.org/edge-lambda/src/wellcomeLibraryRedirect.ts
kenoir: > Possibly move the extraction logic into a separate file? It's just string parsing, could be managed/tested on its own. 

👍

alexwlchan: > I'm not sure if they can in this context, this is redirecting a specific pattern from the wellcomelibrary site.

If so, maybe better to match on 7 digits? 8 digits implies the presence of a check digit.

jamesgorrie: I would also say it's odd practise in terms of it not really being errors, but rather known states of paths we would redirect.

The more common approach would be to return either `undefined` or a reason if that's needed later for logging, but you could just do the logging here.

cloudfront/wellcomelibrary.org/edge-lambda/src/wellcomeLibraryRedirect.ts
jamesgorrie: [Assuming that this is saying to deal with this more gracefully](https://github.com/wellcomecollection/platform-infrastructure/pull/116/files#diff-cad288977e3edeafbbc2ad132d4fd147a6762ba2623c35f9fc6379e3d638d2d3R38)

I would also say it's odd practise in terms of it not really being errors, but rather known states of paths we would redirect.

The more common approach would be to return either undefined or a reason if that's needed later for logging, but you could just do the logging here.

jamesgorrie (commented):
small suggestion on dealing with asserts/

cloudfront/wellcomelibrary.org/edge-lambda/src/catalogue.ts
```diff
@@ -0,0 +1,14 @@
...
+    id: string;
+    title: string;
+};
+
+export type CatalogueResultsList<Result = Work> = {
+    type: 'ResultList';
+    totalResults: number;
+    results: Result[];
+    pageSize: number;
+    prevPage: string | null;
```
jamesgorrie, line 12: This is `string | undefined`, as is `nextPage`

kenoir: Literally copied out of https://github.com/wellcomecollection/wellcomecollection.org/blob/master/common/model/catalogue.ts#L299

cloudfront/wellcomelibrary.org/edge-lambda/src/wellcomeLibraryRedirect.ts
kenoir: I couldn't find any links of that form (7 digits), but I don't think there's any harm in being flexible (and possibly some in not).

cloudfront/wellcomelibrary.org/edge-lambda/src/api.test.ts
```diff
@@ -0,0 +1,23 @@
+import axios from 'axios';
+import {getWork} from "./api";
+import {testDataNoResults, testDataSingleResult} from './apiFixtures'
```
jamieparkinson, line 3: This is a nit but you may want to get `prettier` autoformatting set up in here (will sort out spacing, single/double quotes, end-of-file newlines, etc etc).

cloudfront/wellcomelibrary.org/edge-lambda/src/api.ts
```diff
@@ -0,0 +1,59 @@
+import axios from "axios";
+import {Work, CatalogueResultsList} from "./catalogue";
+
+export type GetWorkResult = Work | Error
+
+const apiBasePath = 'https://api.wellcomecollection.org/catalogue/v2'
+
+type ApiQuery = {
+    query: string
+    include: Array<string>
```
jamesgorrie, line 10: style: `string[]`
we can setup some linting rules too.

cloudfront/wellcomelibrary.org/edge-lambda/src/api.ts
```diff
@@ -0,0 +1,59 @@
...
+}
+
+async function* apiQuery(query: ApiQuery): AsyncGenerator<Work, void, void> {
+    const url = apiBasePath + '/works';
```
jamesgorrie, line 14: style:
```suggestion
    const url = `${apiBasePath}/works`;
```

cloudfront/wellcomelibrary.org/edge-lambda/src/api.ts
```diff
@@ -0,0 +1,59 @@
...
+    const apiResult = await axios.get(url, {params: query}  );
```
jamesgorrie, line 15: If you call the argument `params` you can just have
```suggestion
    const apiResult = await axios.get(url, { params });
```

I personally really like to keep arguments between methods the same to avoid cognitive load on what's going on. With limits obviously.

cloudfront/wellcomelibrary.org/edge-lambda/src/api.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    const resultList = apiResult.data as CatalogueResultsList
+    // const nextPage = resultList.nextPage
+    //
+    // if(resultList.results.length == 0) {
+    //     return Error("No matching Catalogue API results found")
+    // }
+
+    for(let i = 0; i < resultList.results.length; i++) {
+        yield resultList.results[i];
```
jamesgorrie, line 25: Have you read this?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of

cloudfront/wellcomelibrary.org/edge-lambda/src/catalogue.ts
```diff
@@ -0,0 +1,14 @@
...
+    id: string;
+    title: string;
+};
+
+export type CatalogueResultsList<Result = Work> = {
+    type: 'ResultList';
+    totalResults: number;
+    results: Result[];
+    pageSize: number;
+    prevPage: string | undefined;
```
jamesgorrie, line 12:
```suggestion
    prevPage?: string
```

There is a slight difference if you are implementing the type. e.g.
```ts
type A = { a: string | undefined }
const { a: undefined } // valid
const {} // invalid

type B = { b?: string }
const { a: undefined } // valid
const {} // valid
```

cloudfront/wellcomelibrary.org/edge-lambda/src/api.ts
kenoir: I did have a look over async generators/iterables, but got in a muddle and decided to come back later. I'm not sure how best to handle providing a generator that does async requests at the end of each page. I also want to be able to filter the output of that generator.

cloudfront/wellcomelibrary.org/edge-lambda/src/api.ts
```diff
@@ -0,0 +1,63 @@
...
+  pageSize: number;
+  prevPage?: string;
+  nextPage: string | undefined;
+};
+
+export type GetWorkResult = Work | Error;
+
+export async function* apiQuery(
+  query: ApiQuery
+): AsyncGenerator<Work, void, void> {
```
jamieparkinson, line 42: Could you explain a little about why you chose to use an async generator here? I've only come across them very rarely in application code (although that doesn't mean it's a bad choice) with the exception of some libraries that lean on them heavily.

cloudfront/wellcomelibrary.org/edge-lambda/src/api.ts
```diff
@@ -0,0 +1,63 @@
...
+  while (nextPage) {
+    const apiResult = await axios.get(nextPage);
+    const resultList = apiResult.data as CatalogueResultsList;
+    nextPage = resultList.nextPage;
+
+    for await (const result of resultList.results) {
+      yield result;
+    }
+  }
+}
```
jamieparkinson, line 63: Why does this get repeated? What about...
```typescript
let nextPage = true
while (nextPage) {
  const apiResult = await axios.get(nextPage);
  const resultList = apiResult.data as CatalogueResultsList;
  nextPage = resultList.nextPage;

  for await (const result of resultList.results) {
    yield result;
  }
}
```

cloudfront/wellcomelibrary.org/edge-lambda/src/bnumberToWork.test.ts
```diff
@@ -0,0 +1,114 @@
...
+
+function addIdentifiersToPage(
+  identifiers: Identifier[],
+  resultList: CatalogueResultsList
+) {
+  const firstResult = resultList.results[0];
+  const resultWithIdentifiers = Object.assign(firstResult, {
+    identifiers: identifiers,
+  });
+  return Object.assign(resultList, { results: [resultWithIdentifiers] });
```
jamieparkinson, line 73: Matter of taste but I would usually use the spread operator rather than `Object.assign`:
```typescript
return {
  ...resultList,
  results: [resultWithIdentifiers]
};
```

cloudfront/wellcomelibrary.org/edge-lambda/src/bnumberToWork.test.ts
```diff
@@ -0,0 +1,114 @@
...
+    type: 'Identifier',
+  } as Identifier,
+];
+
+function addIdentifiersToPage(
+  identifiers: Identifier[],
+  resultList: CatalogueResultsList
+) {
+  const firstResult = resultList.results[0];
+  const resultWithIdentifiers = Object.assign(firstResult, {
```
jamieparkinson, line 70: Nit/taste but you can use spread + property shorthands:
```typescript
const resultWithIdentifiers = { ...firstResult, identifiers };
```

cloudfront/wellcomelibrary.org/edge-lambda/src/bnumberToWork.ts
```diff
@@ -0,0 +1,34 @@
...
+      // If the work has a sierra-identifier identifier, that
+      // preferentially identifiers the work as being sourced
+      // from Sierra, so use that work if we see it.
+      const filteredIdentifier = identifiers.filter(
+        (thing) => thing.identifierType.id === 'sierra-identifier'
+      );
+
+      if (filteredIdentifier.length >= 1) {
+        break;
+      }
```
jamieparkinson, line 25: I think you can express this more directly:
```typescript
const hasSierraId = identifiers.some(
  thing => thing.identifierType.id === 'sierra-identifier'
);
if (hasSierraId) {
  break;
}
```

jamieparkinson: In fact you might be able to rewrite this function without the mutation & break:
```typescript
const resultList = apiQuery({
  query: bNumber,
  include: ['identifiers'],
});

for await (const result of resultList) {
  const identifiers: Identifier[] = result.identifiers;
  const hasSierraId = identifiers.some(
    thing => thing.identifierType.id === 'sierra-identifier'
  );
  if (hasSierraId) {
    return result;
  }
}

return Error('No matching Catalogue API results found');
```

cloudfront/wellcomelibrary.org/edge-lambda/src/paths.test.ts
```diff
@@ -0,0 +1,63 @@
+import { expect, test } from '@jest/globals';
+import { getBnumberFromPath } from './paths';
+
+type ExpectedPath = {
+  in: string;
+  out: string | Error;
+};
+const pathTests = (): Array<ExpectedPath> => {
```
jamieparkinson, line 8: nit, prefer `ExpectedPath[]`

cloudfront/wellcomelibrary.org/edge-lambda/src/paths.ts
```diff
@@ -0,0 +1,28 @@
+export type GetBNumberResult = string | Error;
```
jamieparkinson, line 1: Boring thought: you could make this look even more like scala by creating a generic for reuse everywhere:
```typescript
type Try<T> = T | Error;
export type GetBNumberResult = Try<string>;
```

cloudfront/wellcomelibrary.org/edge-lambda/src/paths.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export function getBnumberFromPath(path: string): GetBNumberResult {
+  const splitPath = path.split('/');
+
+  // Match on paths like b1234567x / b12345678
+  const bNumberRegexp = /^[bB][0-9]{7}/;
+
+  if (splitPath[0] !== '') {
+    return Error(`Path ${path} does not start with /`);
```
jamieparkinson, line 10: Big fan of flat guard clauses like this, by far the clearest way to represent this kind of logic 👍

cloudfront/wellcomelibrary.org/edge-lambda/src/redirectToRoot.ts
```diff
@@ -0,0 +1,13 @@
+import { CloudFrontRequest } from 'aws-lambda/common/cloudfront';
+import { createRedirect } from './createRedirect';
+
+export function redirectToRoot(request: CloudFrontRequest) {
+  if (request.headers.host && request.headers.host.length === 1) {
```
jamieparkinson, line 5:
```typescript
request.headers.host?.length === 1
```

cloudfront/wellcomelibrary.org/edge-lambda/src/wellcomeLibraryRedirect.ts
```diff
@@ -1,21 +1,62 @@
...
-    const requestHost =  request.headers.host[0].value
+    // Find corresponding work id
+    const bNumber = bNumberResult;
+    const work = await getWork(bNumber);
 
-    if (requestHost.startsWith('www.')) {
-      const rootRequestHost = requestHost.replace('www.','');
-      return Promise.resolve(redirect(`https://${rootRequestHost}${request.uri}`));
+    if (work instanceof Error) {
+      return notFoundRedirect;
```
jamieparkinson, line 33: Will there be a way of distinguishing which of these 2 error cases happened?

jamieparkinson (approved):
I think this is good - I've made some style suggestions and am interested in the async generator stuff

cloudfront/wellcomelibrary.org/edge-lambda/package-lock.json
```diff
@@ -0,0 +1,6289 @@
+{
```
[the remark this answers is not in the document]
kenoir: I think I inadvertently run an `npm` command instead of `yarn` and carelessly committed this.

cloudfront/wellcomelibrary.org/edge-lambda/src/paths.ts
kenoir: I did think about that, but ended up not as I feel it leads towards "use a library for functional stuff" which i'd like to avoid here at least.

cloudfront/wellcomelibrary.org/edge-lambda/src/wellcomeLibraryRedirect.ts
kenoir: We may add metrics later, I should at least log these errors I think - will fix.

cloudfront/wellcomelibrary.org/edge-lambda/src/api.ts
kenoir:
```ts
let nextPage = true
while (nextPage) {
  // nextPage is `true` in the first loop and doesn't work
  // I've tried a couple of ways and nothing I found is more 
  // satisfying than just putting up with the duplication
  const apiResult = await axios.get(nextPage);
  const resultList = apiResult.data as CatalogueResultsList;
  nextPage = resultList.nextPage;

  for await (const result of resultList.results) {
    yield result;
  }
}
```

cloudfront/wellcomelibrary.org/edge-lambda/src/api.ts
kenoir: It's not really a carefully considered choice. Thinking "I need something like a generator here, and the underlying requests should be async", cue me googling "javascript async generator".

cloudfront/wellcomelibrary.org/edge-lambda/src/api.ts
jamieparkinson: ohhhh I see, yes, fair enough

cloudfront/wellcomelibrary.org/edge-lambda/src/bnumberToWork.ts
kenoir: I've a hard time trying to make this work without mutation while trying to keep the leniency about the existence of identifiers & fallback to record without matching a sierra identifier identifier. The use of `some` here though is a revelation.

PR #117, opened 2021-03-03
Update IIIF/DLCS Cloudfront rules for test + staging

donaldgray:
Summary of changes:

* Update origin id `"dds"` -> `"dashboard"`.
* Update dashboard + iiif origin domains from `*.dlcs.io` -> `*.wellcomecollection.digirati.io`
* Add new origin for pdf-coverpage
* Switched order of thumbs behaviours to ensure `/b*` is before `/*`
* Update lambda for AV 
* Forward host header to AV/image + thumbs as this'll be needed for setting id's in some return result (will set this in "dlcs_wellcome_images" behaviours once tested.

Unfortunately don't have appropriate rights to `terraform plan` locally so let me know if there's any issues.


cloudfront/iiif.wellcomecollection.org/terraform/cf_behaviours.tf
```diff
@@ -196,35 +248,58 @@ locals {
     },
   ]
 
+  pdf_cover_behaviours = [
+    {
+      path_pattern     = "pdf-cover/*"
+      target_origin_id = "pdf_cover"
+      headers          = []
+      cookies          = null
```
kenoir, line 256: This can't be `null`.

cloudfront/iiif.wellcomecollection.org/terraform/cf_behaviours.tf
```diff
@@ -100,7 +100,7 @@ locals {
     {
       path_pattern     = "image/*"
       target_origin_id = "dlcs_images"
-      headers          = []
+      headers          = ["Host"]
```
kenoir, line 103: When testing this the addition of the `Host` header forwarding seems to break the behaviour entirely!

cloudfront/iiif.wellcomecollection.org/terraform/cf_origins.tf
```diff
@@ -54,18 +54,23 @@ locals {
       origin_id   = "dlcs_file"
       domain_name = "dlcs.io"
       origin_path = "/file/wellcome/5"
+    },
+    {
+      origin_id   = "pdf_cover"
+      domain_name = "pdf.wellcomecollection.digirati.io"
```
kenoir, line 60: CloudFront gets upset trying to apply the prod configuration saying "not all the origins exist", I think it might be that this origin is not contactable.

cloudfront/iiif.wellcomecollection.org/terraform/cf_behaviours.tf
donaldgray: Odd - reverted to `[]`. I added `"Host"` thinking I'll need to see the original host in dlcs to update some returned values based on whether the request was made directly or via CF. I can test without `"Host"` and troubleshoot adding back in if need be.

cloudfront/iiif.wellcomecollection.org/terraform/cf_behaviours.tf
donaldgray: Made `"All"` - pdf_cover page will never care about cookies hence the null

kenoir: I think you could make this "none".

PR #119, opened 2021-03-08
No change wellcomelibrary.org passthrough

kenoir:
Simplest no-change update to redirect requests via CloudFront.

This change removes any extras and establishes a baseline no-change switcheroo for requests coming via wellcomelibrary.org that introduces CloudFront as an intermediary.

Note: This change is applied.


cloudfront/wellcomelibrary.org/terraform/dns.tf
```diff
@@ -1,9 +1,45 @@
...
+  type    = "A"
+
+  weighted_routing_policy {
+    weight = 50
+  }
+
+  alias {
+    name                   = "d2cksn8op995w4.cloudfront.net"
+    evaluate_target_health = false
+    zone_id                = "Z2FDTNDATAQYW2"
```
alexwlchan, line 29: Where does this Zone ID come from?

cloudfront/wellcomelibrary.org/terraform/dns.tf
```diff
@@ -1,9 +1,45 @@
...
+  zone_id = data.aws_route53_zone.zone.id
+  name    = "wellcomelibrary.org"
+  type    = "A"
+
+  weighted_routing_policy {
+    weight = 50
+  }
+
+  alias {
+    name                   = "d2cksn8op995w4.cloudfront.net"
```
alexwlchan, line 27: Where does this CloudFront name come from?

cloudfront/wellcomelibrary.org/edge-lambda/src/redirect.ts
```diff
@@ -0,0 +1,14 @@
+import {CloudFrontResultResponse} from "aws-lambda/common/cloudfront";
+
+export async function redirect(uri: string){
+    return {
+        status: '301',
```
alexwlchan, line 5: Two minor thoughts:

* I wonder if we should stick to 302 until we've got the redirection stuff sorted, so clients don't end up caching the redirects?
* Why is this function async?

alexwlchan (approved):

cloudfront/wellcomelibrary.org/edge-lambda/src/redirect.ts
kenoir: - Good idea on 302.
- The function need not be async.

cloudfront/wellcomelibrary.org/terraform/dns.tf
kenoir: Good catch - I copy/pasted this to satisfy a terraform import and didn't update it. I can just pull this from the prod CloudFront distro module 👍

cloudfront/wellcomelibrary.org/terraform/dns.tf
kenoir: Same issue as with zone_id.

cloudfront/wellcomelibrary.org/terraform/dns.tf
kenoir: Oh, this is an interesting one actually - this value is the same for all cloudfront distros (added a comment with link to docs).

PR #122, opened 2021-03-10
Add x-forwarded-host to specified origins

donaldgray:
> I can't plan this locally so syntax could be wrong!

Added `forward_host` variable to origin_set. If this is true then `x-forwarded-host` is added to origin custom headers, set to value of alias for distro. Added to `/thumbs/` requests only for now, this will allow me to validate it works without affecting anything called by external code.

Applying this should update the 3 "dlcs_thumbs" origins in the 3 iiif.wc.org distros.

Tested setting this in a test CF distro with and without going via an Origin request lambda and the value is persisted.


kenoir (approved):

kenoir:
This is applied.

PR #123, opened 2021-03-10
Manage the logging cluster in Terraform; connect to it using a VPC endpoint

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5002

I didn't mean this to be quite so disruptive, but in for a penny, in for a pound.

This pull request adds a new Elasticsearch cluster for logging (`ec_deployment.logging`). It also adds traffic filters that allow it to receive traffic from every VPC into which we copy our shared logging secrets. These traffic filters are connected to VPC endpoints, which means logs can be sent cheaper than going over a managed NAT Gateway. It's a big diff, but a lot of it is repetition.

Next steps:

- [x] Get this cluster working with OIDC
- [x] Put this cluster behind logging.wellcomecollection.org
- [x] Set up index expiry for these logs, so we don't fill up the cluster's disk


critical/back_end/modules/elasticsearch_privatelink/provider.tf
```diff
@@ -0,0 +1,11 @@
+# Although we've declared the provider at the top level, we need to redeclare it in
```
kenoir, line 1: Useful comment.

kenoir (approved):
Phew! This is a big change - you've got it out neatly and v.quick too.

PR #125, opened 2021-03-15
Add library API urls

kenoir:
For https://github.com/wellcomecollection/platform/issues/5081


cloudfront/wellcomelibrary.org/edge-lambda/src/wellcomeLibraryRedirect.test.ts
```diff
@@ -61,6 +69,57 @@ const rewriteTests = (): ExpectedRewrite[] => {
...
+    {
+      in: '/iiif/collection/b18031900',
+      out: expectedServerError('Got 404 from https://iiif.wellcomecollection.org/wlorgp/iiif/collection/b18031900'),
+      error: axios404
+    },
+    {
+      in: '/iiif/collection/b18031900',
+      out: expectedServerError('No response from https://iiif.wellcomecollection.org/wlorgp/iiif/collection/b18031900'),
+      error: axiosNoResponse
+    },
```
alexwlchan: Very minor: it wasn't immediately clear to me where these URLs came from, and whether these test cases had been fleshed out properly. Picking separate b numbers for the tests might have made that clearer (and be easier to debug in tests.)

cloudfront/wellcomelibrary.org/edge-lambda/src/wlgorpLookup.ts
```diff
@@ -0,0 +1,23 @@
...
+
+function validateUrl(url: string): Error | URL {
+    try {
+        return new URL(url);
+    } catch (_) {
+        return Error(`Invalid URL: ${url}`);
+    }
+}
+
+export async function wlgorpLookup(lookupUri: string): Promise<Error | URL> {
```
alexwlchan, line 13:
```suggestion
export async function wlorgpLookup(lookupUri: string): Promise<Error | URL> {
```

cloudfront/wellcomelibrary.org/edge-lambda/src/safeGet.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export async function safeGet<T>(
+  url: string,
+  config?: AxiosRequestConfig
+): Promise<Error | T> {
+  const apiResult: AxiosResponse<any> | Error = await axios
+    .get(url, config)
+    .catch((error) => {
+      // Log underlying error here, we don't want to expose it to callers.
+      console.error(error);
```
alexwlchan, line 11: Do you need to log the URL that was associated with this here, or will we be able to match it with the Error block further down?

cloudfront/wellcomelibrary.org/edge-lambda/src/safeGet.ts
```diff
@@ -0,0 +1,27 @@
...
+      }
+
+      return Error(`Unknown error from ${url}: ${error}`);
+    });
+
+  if (apiResult instanceof Error) {
+    return apiResult;
+  }
+
+  return apiResult.data;
```
alexwlchan, line 26: Is this going to bin the response code? Do we care? e.g. a 302 redirect

cloudfront/wellcomelibrary.org/edge-lambda/src/safeGet.ts
kenoir: Ah a trick question! Yes - we should log the URL - good spot.

cloudfront/wellcomelibrary.org/edge-lambda/src/safeGet.ts
kenoir: Yes, it will, but I think in this function we are only interested in responses that return data and the status isn't important. I believe that axios will be default [follow redirects](https://github.com/axios/axios/blob/8b0f373df0574b7cb3c6b531b4092cd670dac6e3/test/unit/adapters/http.js#L98).

jamieparkinson (approved):

PR #126, opened 2021-03-15
Static Wellcome Library redirects from file

kenoir:
Keep the list of static redirects in a user-accessible format i.e. CSV.

Part of https://github.com/wellcomecollection/platform/issues/5080


cloudfront/wellcomelibrary.org/edge-lambda/src/wellcomeLibraryRedirect.test.ts
```diff
@@ -120,3 +128,9 @@ test('leaves other headers unmodified', async () => {
     ],
   });
 });
+
+// test('flarp', async () => {
+//   const foo = await readStaticRedirects()
+//
+//   expect(foo).toBe(true)
+// });
```
jamesgorrie, line 136: Bloopy 👀

cloudfront/wellcomelibrary.org/edge-lambda/src/staticRedirects.ts
```diff
@@ -0,0 +1,26 @@
...
+            .on("error", reject)
+            .on("data", (row: RedirectRow) => {
+                const lookupKey = row.libraryUrl.replace('wellcomelibrary.org', '')
+                redirects[lookupKey] = row.collectionUrl
+            })
+            .on("end", () => {
+                resolve(redirects);
+            });
+    });
+}
```
jamesgorrie, line 26: Instead of us having to read this CSV file, we could just parse it into a ts file and include it [like this](https://github.com/wellcomecollection/wellcomecollection.org/blob/master/cache/edge_lambdas/src/redirects.ts)? Makes it way easier to validate the file in PRs etc, and ensures it's of a standard shape via types.

Just wondering if the data is faulty (returns a borked URL) - where do we deal with that? Or will we just redirect you to something that doesn't exist?

jamesgorrie: Having this as typescript would also catch errors like duplicate keys.

kenoir: I think we should keep the CSV for ease of reading by non devs, but also generate a TS file for code validation and speed in the lambda. We should check in the build that the TS/CSV are synced.

jamesgorrie: Could we perhaps has a TS file that generates the CSV? Eventually it will be devs predominantly adding to this, and editing CSVs is annoying and error prone.

A complete nit, and non-blocking, as we can change it later.

jamesgorrie: I do think we should get the CSV parsing out of the lambda though.

jamesgorrie: Interesting read on this: https://aws.amazon.com/blogs/networking-and-content-delivery/leveraging-external-data-in-lambdaedge/

kenoir: > Global variables that you declare outside of the function handler persist across multiple Lambda@Edge invocations.
😱

kenoir: >Could we perhaps has a TS file that generates the CSV? Eventually it will be devs predominantly adding to this, and editing CSVs is annoying and error prone.

The whole point is to make this readable / roughly editable by folk who don't want to read typescript, and might pass us lists of links in volume (so a csv->ts conversion is required anyway) so I think we should keep a CSV as a source.

cloudfront/wellcomelibrary.org/edge-lambda/package.json
```diff
@@ -7,15 +7,17 @@
...
+    "lint": "eslint src/**/*.ts --fix && eslint *.ts --fix",
+    "build": "tsc -p tsconfig.build.json && yarn bundle ",
     "bundle": "cp package.json dist && cd dist && yarn install --production && zip -r wellcome_library_redirect.zip . && cd ..",
     "deploy": "yarn build && yarn test && node deploy",
     "dockerLoginLocal": "aws ecr get-login-password --region eu-west-1 --profile platform | docker login --username AWS --password-stdin 760097843905.dkr.ecr.eu-west-1.amazonaws.com",
     "dockerBuildLocal": "yarn dockerLoginLocal && docker build . -t weli_cf_lambdas",
     "dockerTestLocal": "yarn dockerBuildLocal && docker run weli_cf_lambdas yarn test",
-    "dockerDeployLocal": "yarn dockerTestLocal && docker run weli_cf_lambdas -v ~/.aws:/root/.aws yarn deploy"
+    "dockerDeployLocal": "yarn dockerTestLocal && docker run weli_cf_lambdas -v ~/.aws:/root/.aws yarn deploy",
+    "generateStaticRedirects": "ts-node generateStaticRedirects.ts",
```
jamieparkinson, line 19: Could you stick this into the build/bundle command so it doesn't get forgotten?

cloudfront/wellcomelibrary.org/edge-lambda/src/staticRedirects.json
```diff
@@ -0,0 +1,29 @@
+{
```
jamieparkinson, line 1: How come we're committing the JSON _and_ the CSV? Couldn't we get away with just the CSV?

cloudfront/wellcomelibrary.org/edge-lambda/src/wellcomeLibraryRedirect.test.ts
```diff
@@ -120,3 +160,9 @@ test('leaves other headers unmodified', async () => {
     ],
   });
 });
+
+// test('flarp', async () => {
```
jamieparkinson, line 164: oop

cloudfront/wellcomelibrary.org/edge-lambda/generateStaticRedirects.js
```diff
@@ -0,0 +1,24 @@
+"use strict";
```
jamieparkinson, line 1: Should this file be committed?

jamieparkinson (changes requested):

cloudfront/wellcomelibrary.org/edge-lambda/generateStaticRedirects.js
```diff
@@ -0,0 +1,24 @@
...
+            var lookupKey = row.libraryUrl.replace('wellcomelibrary.org', '');
+            redirects[lookupKey] = row.collectionUrl;
+        })
+            .on("end", function () {
+            resolve(redirects);
+        });
+    });
+}
+exports.readStaticRedirects = readStaticRedirects;
+console.log('foo');
```
jamesgorrie, line 24: Looks like we're doing this in TS and JS?

cloudfront/wellcomelibrary.org/edge-lambda/src/staticRedirects.json
jamesgorrie: I suppose you'd just have to make sure you build the assets as part-of or before the test phase.

jamieparkinson: I would just whack it in the test/build commands and add it to the .gitignore...

cloudfront/wellcomelibrary.org/edge-lambda/generateStaticRedirects.js
kenoir: narp

cloudfront/wellcomelibrary.org/edge-lambda/package.json
kenoir: I deliberately did not do it as part of the build, and had the test step verify that it had been built. I can't remember why now, generating it in the build is better - i've done that.

jamieparkinson (approved):

PR #137, opened 2021-03-25
Add tests for Wellcome Library API redirects

kenoir:
Part of https://github.com/wellcomecollection/platform/issues/5081


cloudfront/wellcomelibrary.org/edge-lambda/apiRedirects.csv
```diff
@@ -0,0 +1,14 @@
+sourceUrl,targetUrl
+wellcomelibrary.org/iiif/b28047345/manifest,https://iiif.wellcomecollection.org/presentation/v2/b28047345
+wellcomelibrary.org/iiif/collection/b18031511,https://iiif.wellcomecollection.org/presentation/v2/b18031511
+wellcomelibrary.org/iiif/b18031511-2/manifest,https://iiif.wellcomecollection.org/presentation/v2/b18031511_0003
+wellcomelibrary.org/item/b18031511,https://wellcomecollection.org/works/ysmqsfhg
```
alexwlchan, line 5: Without having looked further – I assume that `ysmqsfhg` is the work corresponding to b18031511.

What if I look up an ID that can't be found in the catalogue API? e.g. an old link to a now-deleted item.

Ditto other URLs, I guess.

cloudfront/wellcomelibrary.org/edge-lambda/apiRedirects.csv
```diff
@@ -0,0 +1,14 @@
+sourceUrl,targetUrl
```
alexwlchan, line 1: Would it be worth an optional "description" field that has a human-readable description of the test case?

cloudfront/iiif.wellcomecollection.org/tests/requirements.txt
```diff
@@ -0,0 +1,2 @@
+click
+requests
```
alexwlchan, line 2: I would recommend renaming this to `requirements.in`, and using `pip-tools` to create a `requirements.txt` that freezes a known-good set of dependencies. That way this has more chance of continuing to work in future, rather than breaking when one of these packages releases a new version.

See https://hynek.me/articles/python-app-deps-2018/#pip-tools--everything-old-is-new-again

cloudfront/iiif.wellcomecollection.org/tests/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM python:3
+
+WORKDIR /usr/local/bin
```
alexwlchan, line 3: Do you need a working directory?

alexwlchan:
I don't understand the TypeScript well enough to comment on that, but I had a few comments on other bits.

cloudfront/wellcomelibrary.org/edge-lambda/apiRedirects.csv
kenoir: Try it: http://stage.wellcomelibrary.org/item/blarg (should 404 you deliberately on a works page).

It might be a bit unclear but these are ~smoke tests to ensure the deployed code is working as expected, so they won't have deliberately failing test cases.

cloudfront/wellcomelibrary.org/edge-lambda/apiRedirects.csv
kenoir: Good idea.

cloudfront/iiif.wellcomecollection.org/tests/requirements.txt
kenoir: I'm familiar, but didn't go to the extra trouble here. It probably is worth it 👍

cloudfront/iiif.wellcomecollection.org/tests/Dockerfile
kenoir: Nope.

cloudfront/iiif.wellcomecollection.org/tests/requirements.txt
kenoir: I had difficulty getting pip-tools to work for me, so i've fixed the versions in requirements.txt.

PR #140, opened 2021-03-26
HTTP to HTTPS redirect with CORS headers

kenoir:
For https://github.com/wellcomecollection/platform/issues/5086


jamesgorrie (approved):
What's the need to use the `CloudFront-Forwarded-Proto` over using the URLs protocol? Or is that not forwarded anywhere bar this header?

PR #142, opened 2021-03-29
Recurse through @id containing elements

kenoir:
This started failing (https://buildkite.com/wellcomecollection/platform-infrastructure-redirects/builds?branch=master&page=7), and i'm not sure why - this change recurses through "service"s in the returned JSON. 

It does look like the results of the API have changed though. e.g.

Previous:
```

Checking: https://iiif-test.wellcomecollection.org/image/b19582183_RAMC_391_4_0001.jp2/info.json
--
  | Found 'https://iiif-test.wellcomecollection.org/image/b19582183_RAMC_391_4_0001.jp2'
  | Found 'https://iiif-test.wellcomecollection.org/auth/clickthrough'
  | Found 'https://iiif-test.wellcomecollection.org/auth/clickthrough/logout'
  | Found 'https://iiif-test.wellcomecollection.org/auth/token'
  | Found 'https://iiif-test.wellcomecollection.org/auth/login'
  | Found 'https://iiif-test.wellcomecollection.org/auth/login/logout'
  | Found 'https://iiif-test.wellcomecollection.org/auth/token'

```
Now:
```
Checking: https://iiif-test.wellcomecollection.org/image/b19582183_RAMC_391_4_0001.jp2/info.json
Found 'https://iiif-test.wellcomecollection.org/image/b19582183_RAMC_391_4_0001.jp2'
Found 'https://iiif-test.wellcomecollection.org/auth/clickthrough/logout'
Found 'https://iiif-test.wellcomecollection.org/auth/token'
```

### DIFF
--- a/cloudfront/iiif.wellcomecollection.org/tests/rewrite_tests.py
+++ b/cloudfront/iiif.wellcomecollection.org/tests/rewrite_tests.py
@@ -44,21 +44,22 @@ def validate_auth(info_json, regex):
             return
 
         service_description = json.get("service", [])
+        if isinstance(service_description, dict):
+            service_description = [service_description]
 
-        if "service" in service_description:
-            for service in service_description["service"]:
-                info_id = service.get("@id", None)
-                if not p.match(info_id):
-                    click.echo(
-                        click.style(
-                            f"Id fail - expected '{regex}' but found '{info_id}'",
-                            fg="red",
-                        )
+        for service in service_description:
+            info_id = service.get("@id", None)
+            if not p.match(info_id):
+                click.echo(
+                    click.style(
+                        f"Id fail - expected '{regex}' but found '{info_id}'",
+                        fg="red",
                     )
-                else:
-                    click.echo(click.style(f"Found '{info_id}'", fg="green"))
+                )
+            else:
+                click.echo(click.style(f"Found '{info_id}'", fg="green"))
 
-                validate_service(service)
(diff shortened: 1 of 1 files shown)


donaldgray:
Configuration change in DLCS let to a change in auth services returned in info.json. There was an unnecessary 2nd auth service output for `clickthrough` resources which was removed.

This broke the tests as they made an assumption that services would always be an array, but due to removal of 2nd service it is now a dict. So, this part is no longer correct

```py
service = json.get("service", [])

for s in service:
```

cloudfront/iiif.wellcomecollection.org/tests/rewrite_tests.py
```diff
@@ -43,20 +43,22 @@ def validate_service(json):
...
-
-        for s in service:
-            info_id = s.get("@id", None)
-            if not p.match(info_id):
-                click.echo(
-                    click.style(
-                        f"Id fail - expected '{regex}' but found '{info_id}'", fg="red"
+        service_description = json.get("service", [])
+
+        if "service" in service_description:
```
donaldgray, line 48: This will result in some checks being missed if there isn't a nested service.

donaldgray (changes requested):
Marking as "Request changes" but I'll push a change shortly as I think there's a simpler fix that'll catch everything..

donaldgray (approved):

PR #143, opened 2021-03-29
Final round of static redirects

kenoir:
Adds the final round of static redirects into prod.


cloudfront/wellcomelibrary.org/terraform/locals.tf
```diff
@@ -4,14 +4,14 @@ locals {
...
-  wellcome_library_redirect_arn_prod = "${local.wellcome_library_redirect_arn}:35"
+  wellcome_library_redirect_arn_prod = "${local.wellcome_library_redirect_arn}:44"
 
   wellcome_library_passthru_arn        = aws_lambda_function.wellcome_library_passthru.arn
   wellcome_library_passthru_latest     = aws_lambda_function.wellcome_library_passthru.version
   wellcome_library_passthru_arn_latest = "${local.wellcome_library_passthru_arn}:${local.wellcome_library_passthru_latest}"
   wellcome_library_passthru_arn_stage  = local.wellcome_library_passthru_arn_latest
   # This should be set manually when a stable prod deploy is established.
-  wellcome_library_passthru_arn_prod = "${local.wellcome_library_passthru_arn}:1"
+  wellcome_library_passthru_arn_prod = "${local.wellcome_library_passthru_arn}:26"
```
kenoir, line 14: These are wrong now - fix these!

cloudfront/wellcomelibrary.org/terraform/cf_behaviours.tf
```diff
@@ -118,6 +118,38 @@ locals {
   ]
 
   static_redirects_behaviours_stage = [
+    {
```
kenoir, line 121: This list is not yet exhasutive.

jamesgorrie: Should we have these as a list to map over?

cloudfront/wellcomelibrary.org/terraform/locals.tf
jamesgorrie: Could you leave a note as to what they are / where they come from?

kenoir: The comment above is supposed to cover it.

>This should be set manually when a stable prod deploy is established

cloudfront/wellcomelibrary.org/terraform/cf_behaviours.tf
kenoir: Relatively soon this will just be everything, so i'm not sure how much value there is in optimising it just to remove it again next week.

PR #153, opened 2021-04-08
Cloudfront invalidation

donaldgray:
## What's changing and why?

In addition to below, added CF distribution id to iiif + api modules to allow these to be referenced below.

Added `/cloudfront/invalidation/` folder which includes:

_/lambda_
Lambda for doing invalidation. Copied layout + commands from other lambdas in same repo. Distribution used for invalidation comes from `DISTRIBUTION_ID` envvar.

Uploaded to `wellcomecollection-platform-infra/lambdas/cloudfront_invalidation/sns_handler.zip`. 

No tests as it's making a CF api request and unsure how to test that.

_/terraform_

Use `sns_lambda` module for setting 4 topic + lambda for invalidating:
* api.wc.org
* api-stage.wc.org
* iiif.wc.org
* iiif-stage.wc.org

IAM permissions are set at this level rather than in module to allow flexibility moving forward. Currently configured to allow role_ids from dashboard to publish to SNS topic.

_/terraform/sns_lambda_

Module for creating SNS topic, lambda function and subscription to associate.

> I'm unable to plan this so there may be obvious typos/errors!


cloudfront/invalidation/lambda/src/cache_invalidation.test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  AWSMock.setSDKInstance(AWS);
+  let called = false;
+  AWSMock.mock("CloudFront", "createInvalidation", (params: CloudFrontInvalidationRequest, callback: Function) => {
+    called = true;
+    callback(null, null);
+  })
+
+  await handler(fakeEvent, {} as Context, () => {});
+  expect(called).toBe(true);
```
jamesgorrie, line 25: Should we be restoring? `AWSMock.restore('CloudFront');`

cloudfront/invalidation/lambda/.prettierrc.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
+  trailingComma: 'es5',
+  singleQuote: true,
+};
```
jamesgorrie, line 4: We should set this universally at the root.

cloudfront/invalidation/lambda/.gitignore
```diff
@@ -0,0 +1,4 @@
+dist/*
+node_modules/*
+yarn-error.log
+.yarn
```
jamesgorrie, line 4: Could we just have a universal one?

cloudfront/invalidation/lambda/src/cache_invalidation.test.ts
donaldgray: Yes we should! Updated.

cloudfront/invalidation/lambda/.gitignore
donaldgray: Maybe one for @kenoir - added as separate as the other lambdas in repo have their own .gitignore. Happy to move or leave

cloudfront/invalidation/lambda/.prettierrc.js
donaldgray: Similar to `.gitignore` reply - should we move `.prettierrc.js` for all?

kenoir: I think that it's not blocking for this PR, though a good idea for a Wellcome dev to do this in a future PR.

cloudfront/invalidation/lambda/.gitignore
kenoir: See other comment about this being non-blocking.

cloudfront/invalidation/lambda/src/cache_invalidation.ts
```diff
@@ -0,0 +1,62 @@
+import { SNSHandler, SNSMessage } from 'aws-lambda/trigger/sns';
```
kenoir, line 1: This file should be called `cacheInvalidation.ts` to keep in line with the other lambdas.

PR #174, opened 2021-05-17
Moh cloudfront

jamesgorrie:
Ref https://github.com/wellcomecollection/platform/issues/5114

## What's changing and why?
* add the MoH origin with relevant path patterns to stage
* remove non-user API/item behaviours

## `terraform plan` diff
Will show when we do the prod version

### DIFF
--- a/cloudfront/wellcomelibrary.org/terraform/cf_behaviours.tf
+++ b/cloudfront/wellcomelibrary.org/terraform/cf_behaviours.tf
@@ -118,7 +118,7 @@ locals {
       target_origin_id = local.wellcome_library_moh_origin.origin_id
       headers          = []
       cookies          = "all"
-      lambdas = []
+      lambdas          = []
 
       min_ttl     = null
       default_ttl = null
@@ -140,7 +140,7 @@ locals {
       target_origin_id = local.wellcome_library_moh_origin.origin_id
       headers          = []
       cookies          = "all"
-      lambdas = []
+      lambdas          = []
 
       min_ttl     = null
       default_ttl = null
--- a/cloudfront/wellcomelibrary.org/terraform/cf_behaviours.tf
+++ b/cloudfront/wellcomelibrary.org/terraform/cf_behaviours.tf
@@ -101,64 +101,12 @@ locals {
     }
   ]
 
+  moh_paths = ["moh/*", "spas/*", "assets/*", "plugins/*", "scripts/*", "timelines/*"]
+
   moh_behaviours = [
+    for path in local.moh_paths:
     {
-      path_pattern     = "moh/*"
-      target_origin_id = local.wellcome_library_moh_origin.origin_id
-      headers          = []
-      cookies          = "all"
-      lambdas          = []
-
-      min_ttl     = null
-      default_ttl = null
-      max_ttl     = null
-    },
-    {
-      path_pattern     = "spas/*"
-      target_origin_id = local.wellcome_library_moh_origin.origin_id
-      headers          = []
-      cookies          = "all"
-      lambdas          = []
-
-      min_ttl     = null
-      default_ttl = null
-      max_ttl     = null
-    },
-    {
-      path_pattern     = "assets/*"
-      target_origin_id = local.wellcome_library_moh_origin.origin_id
-      headers          = []
-      cookies          = "all"
-      lambdas          = []
-
-      min_ttl     = null
-      default_ttl = null
-      max_ttl     = null
-    },
-    {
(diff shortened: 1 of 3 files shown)


alexwlchan:
Why do some of the paths hit a Lambda, and some not?

jamesgorrie:
> Why do some of the paths hit a Lambda, and some not?

Because I copy/pasted badly. It's been removed.

cloudfront/wellcomelibrary.org/terraform/cf_behaviours.tf
```diff
@@ -101,7 +101,78 @@ locals {
...
+      target_origin_id = local.wellcome_library_moh_origin.origin_id
+      headers          = []
+      cookies          = "all"
+      lambdas          = []
+
+      min_ttl     = null
+      default_ttl = null
+      max_ttl     = null
+    }
+  ]
```
alexwlchan, line 171: Given these are all the same, you could use a [for expression](https://www.terraform.io/docs/language/expressions/for.html):

```suggestion
  moh_paths = ["moh/*", "spas/*", "assets/*", "plugins/*", "scripts/*", "timelines/*"]

  moh_behaviours = [
    for path in local.moh_paths:
    {
      path_pattern     = path
      target_origin_id = local.wellcome_library_moh_origin.origin_id
      headers          = []
      cookies          = "all"
      lambdas          = []

      min_ttl     = null
      default_ttl = null
      max_ttl     = null
    }
  ]
```

alexwlchan (approved):
LGTM 👍

cloudfront/wellcomelibrary.org/terraform/cf_behaviours.tf
```diff
@@ -101,7 +101,78 @@ locals {
...
+      headers          = []
+      cookies          = "all"
+      lambdas          = []
+
+      min_ttl     = null
+      default_ttl = null
+      max_ttl     = null
+    },
+    {
+      path_pattern     = "assets/*"
```
kenoir: I think this will redirect any existing library pages that use these assets to the MOH version of these assets. There are some of those around (the catalogue pages do for sure). I'm not sure of the best way to resolve this issue - you may need to go back to @tomcrane and ask for the MOH pages to refer to any of these shared resources on /moh/:path:

tomcrane: All the assets from wl.org are in this app. It's doing double duty as the stand-in for wl.org referenced static content/page furniture, as well as the MOH functionality.

Something has to do this, otherwise embedded wl.org content will break when the Euston Rd wl.org shuts down (imminently).

tomcrane: Initially I did move everything under /moh/ until I realised that something had to do this job, and I realised that this app would continue to respond to wl.org, so it might as well be this app, which makes heavy use of those same assets itself.

kenoir: For an example have a look at what https://catalogue.wellcomelibrary.org/ is loading. It's looking at /assets for sure - if the version of /assets on the MOH endpoint is different in some way from that on wellcomelibrary.org it might cause an issue.

jamesgorrie: We've had the chat, here's the response:

> James Gorrie:
> Great.
> Just wondering - we have some references to paths that are mentioned in this issue.
> e.g. https://search.wellcomelibrary.org
> Has a reference to https://wellcomelibrary.org/assets/img/wellcome-logo.png
> Which is OK, as that's available from the new MoH service.
> But http://archives.wellcomelibrary.org has way more references, if I swap over all /assets is that going to break these > older sites?

> Tom Crane:
> I copied all of the wl.org assets in there
> should be OK

jamesgorrie: Just confirming @kenoir and @tomcrane - it's desired that this new app serves everything that makes requests to `/assets` and no changes are necessary?

tomcrane: @jamesgorrie yes, that's the intent.

kenoir: Confirming the new MOH site serves everything indicated by the behaviours changed here. If so and the intent is to move hosting of the static assets to the MOH app (which will eventually be the only place they are referenced then good to go.

PR #178, opened 2021-05-19
Move s3 redirect at API root to api.wellcomecollection.org

kenoir:
## What's changing and why?

The S3 bucket & object in `critical/user_facing` provide the redirect at the base of `api.wellcomecollection.org` to `developers.wellcomecollection.org`. This should sit next to the infra where it's used.

## `terraform plan` diff

The state has been removed from the `critical/user_facing` stacks.

This _should_ be a no-op change but I think the way policies is described in TF has changed _and_ aws_s3_bucket_object does not support import (so might need to be recreated).

```
# aws_s3_bucket.public_api will be updated in-place

  ~ resource "aws_s3_bucket" "public_api" {
      + acl                         = "public-read"
      + force_destroy               = false
        id                          = "wellcomecollection-public-api"
      + policy                      = jsonencode(
            {
              + Statement = [
                  + {
                      + Action    = "s3:GetObject"
                      + Effect    = "Allow"
                      + Principal = {
                          + AWS = "*"
                        }
                      + Resource  = "arn:aws:s3:::wellcomecollection-public-api/*"
                      + Sid       = "PublicReadForGetBucketObjects"
                    },
                ]
              + Version   = "2008-10-17"
            }
        )
        tags                        = {}
        # (9 unchanged attributes hidden)

      - grant {
          - permissions = [
              - "READ",
            ] -> null
          - type        = "Group" -> null
          - uri         = "http://acs.amazonaws.com/groups/global/AllUsers" -> null
        }
      - grant {
          - id          = "711c7bddce89222cf7830990b7f0c4c2e8a6ba323b5a2fe4168bfd69e19f1e72" -> null
          - permissions = [
              - "FULL_CONTROL",
            ] -> null
          - type        = "CanonicalUser" -> null
        }


        # (2 unchanged blocks hidden)
    }

  # aws_s3_bucket_object.object will be created
  + resource "aws_s3_bucket_object" "object" {
      + acl                    = "private"
      + bucket                 = "wellcomecollection-public-api"
      + content_type           = "text/html"
      + etag                   = "1926c572db8e909135f7cef9d15b558c"
      + force_destroy          = false
      + id                     = (known after apply)
      + key                    = "index.html"
      + kms_key_id             = (known after apply)
      + server_side_encryption = (known after apply)
      + source                 = "./s3_objects/index.html"
      + storage_class          = (known after apply)
      + version_id             = (known after apply)
    }
```


kenoir:
Have tested this and it works.

PR #180, opened 2021-05-19
Remove admin cidr

kenoir:
## What's changing and why?

Remove the reference to admin CIDR in the user_facing/critical stack

## `terraform plan` diff

No-op change.


kenoir:
A search makes it look like monitoring is the only place this is used: https://github.com/search?p=3&q=org%3Awellcomecollection+admin_cidr_ingress&type=Code

alexwlchan (approved):

PR #189, opened 2021-05-27
Output the logging cluster ID and freshen the ES infra

kenoir:
## What's changing and why?

This PR outputs the logging cluster ID so that it can be consumed by the catalogue-pipeline and catalogue-api ES config.

In addition:
- Adds default tags to all resources in the stack
- Provision as much of the logging cluster ES config as we can here.


kenoir:
There is a lot of implicit config in getting a cluster configured for access via Azure AD - this goes a bit further to making all of that config explicit.

alexwlchan (approved):
The ability to configure this in the Elastic Cloud provider is super nice.

kenoir:
For: https://github.com/wellcomecollection/platform/issues/5184

PR #196, opened 2021-06-09
Archive redirects

jamesgorrie:
## What's changing and why?
Ref: https://github.com/wellcomecollection/platform/issues/5058

Adds the redirects for `archives.wellcomecollection.org`.

The DServe application is a little fiddly, but there is some decernable information we can get from the QS.

In short
* `dsqItem` - which has a RefNo in it
* `dsqSearch`- can hold multiple types of search DSL. All search terms that are searched for are wrapped in `'`. We don't bother checking the field etc, we just take all the terms in `'`, join them on ` `, and search for it. If we find a matching ID in a work, we redirect to that work. If we don't, we search for the new search terms.

This should catch most URLs out there. The examples are taken from GA and Google searches, so real things that are happening.

## `terraform plan` diff
No-op.

### DIFF
--- a/cloudfront/wellcomelibrary.org/edge-lambda/src/catalogueApi.ts
+++ b/cloudfront/wellcomelibrary.org/edge-lambda/src/catalogueApi.ts
@@ -36,7 +36,7 @@ export async function* apiQuery(
   query: string
 ): AsyncGenerator<Work, void, void> {
   const url = `${apiBasePath}/works`;
-  const queryUrl = `${url}?include=identifiers&identifiers=${query}`;
+  const queryUrl = `${url}?include=identifiers&query=${query}`;
 
   const apiResult = await axios.get(queryUrl);
   const resultList = apiResult.data as CatalogueResultsList;
--- a/cloudfront/wellcomelibrary.org/edge-lambda/src/catalogueApi.ts
+++ b/cloudfront/wellcomelibrary.org/edge-lambda/src/catalogueApi.ts
@@ -56,3 +56,25 @@ export async function* apiQuery(
     }
   }
 }
+
+export async function findWorkWithIdentifierValue(
+  identifierValue: string,
+  identifierType?: string
+): Promise<Work | undefined> {
+  const resultList = apiQuery(identifierValue);
+
+  for await (const result of resultList) {
+    if (result.identifiers) {
+      const identifiers: Identifier[] = result.identifiers;
+      const hasMatchingId = identifiers.some(
+        (identifier) =>
+          identifier.value.toLowerCase() === identifierValue.toLowerCase() &&
+          (!identifierType || identifier.identifierType.id === identifierType)
+      );
+
+      if (hasMatchingId) {
+        return result;
+      }
+    }
+  }
+}
--- a/cloudfront/wellcomelibrary.org/edge-lambda/src/catalogueApi.ts
+++ b/cloudfront/wellcomelibrary.org/edge-lambda/src/catalogueApi.ts
@@ -68,8 +68,9 @@ export async function findWorkWithIdentifierValue(
       const identifiers: Identifier[] = result.identifiers;
       const hasMatchingId = identifiers.some(
         (identifier) =>
-          identifier.value.toLowerCase() === identifierValue.toLowerCase() &&
-          (!identifierType || identifier.identifierType.id === identifierType)
+          (!identifierType ||
+            identifier.identifierType.id === identifierType) &&
+          identifier.value.toLowerCase() === identifierValue.toLowerCase()
       );
 
       if (hasMatchingId) {
--- a/cloudfront/wellcomelibrary.org/edge-lambda/src/catalogueApi.ts
+++ b/cloudfront/wellcomelibrary.org/edge-lambda/src/catalogueApi.ts
@@ -36,7 +36,7 @@ export async function* apiQuery(
   query: string
 ): AsyncGenerator<Work, void, void> {
   const url = `${apiBasePath}/works`;
-  const queryUrl = `${url}?include=identifiers&query=${query}`;
+  const queryUrl = `${url}?include=identifiers&identifiers=${query}`;
 
   const apiResult = await axios.get(queryUrl);
   const resultList = apiResult.data as CatalogueResultsList;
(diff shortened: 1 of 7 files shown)


cloudfront/wellcomelibrary.org/edge-lambda/src/catalogueApi.ts
```diff
@@ -56,3 +56,25 @@ export async function* apiQuery(
     }
   }
 }
+
+export async function findWorkWithIdentifierValue(
+  identifierValue: string,
+  identifierType?: string
+): Promise<Work | undefined> {
+  const resultList = apiQuery(identifierValue);
```
alexwlchan, line 64: If you're filtering on an `identifier` value, you might want to use the `identifiers` filter.

cloudfront/wellcomelibrary.org/edge-lambda/src/wellcomeLibraryArchiveRedirect.ts
```diff
@@ -1,15 +1,53 @@
...
+  const dsqItem = qs.dsqItem ? qs.dsqItem.toString() : undefined;
+  const dsqSearch = qs.dsqSearch ? qs.dsqSearch.toString() : undefined;
+  const dsqSearchTermsMatch = dsqSearch
+    ? dsqSearch.match(/'([^']*)'/g)
+    : undefined;
+  // The dserve app puts all test it's searching for in the format of `(Field='{term}')`
+  // so we just search for anything between `'`
+  const dsqSearchTerms = dsqSearchTermsMatch
+    ? dsqSearchTermsMatch.map((term) => term.replace(/'/g, '')).join(' ')
+    : undefined;
```
alexwlchan, line 30: What if you search for a string that contains a single quote? Should these only be removed from the start/end?

cloudfront/wellcomelibrary.org/edge-lambda/src/catalogueApi.ts
```diff
@@ -56,3 +56,25 @@ export async function* apiQuery(
...
+
+  for await (const result of resultList) {
+    if (result.identifiers) {
+      const identifiers: Identifier[] = result.identifiers;
+      const hasMatchingId = identifiers.some(
+        (identifier) =>
+          identifier.value.toLowerCase() === identifierValue.toLowerCase() &&
+          (!identifierType || identifier.identifierType.id === identifierType)
```
jamieparkinson, line 72: Since we're in an edge lambda please forgive this perf-optimisation bicycle storage solution: swapping the order of these 2 predicates (the value check and the type check) is marginally preferable because the value check is more computationally expensive

cloudfront/wellcomelibrary.org/edge-lambda/src/wellcomeLibraryArchiveRedirect.ts
jamesgorrie: This get's mangled by the DServe search, and I think that's OK
http://archives.wellcomelibrary.org/DServe/dserve.exe?dsqIni=Dserve.ini&dsqApp=Archive&dsqCmd=Show.tcl&dsqDb=Catalog&dsqPos=0&dsqSearch=%28%28text%29%3D%27-things-%27%29

jamesgorrie: In this instance we will redirect to `query=-things-` which is fine.

cloudfront/wellcomelibrary.org/edge-lambda/src/catalogueApi.ts
jamesgorrie: Any benchmarks?

jamesgorrie: There we have it
https://jsben.ch/bofnA

cloudfront/wellcomelibrary.org/edge-lambda/src/catalogueApi.ts
jamesgorrie: This is copy/pasted, but a nice idea, let me have a go.

jamesgorrie: I've left this as is as it's used elsewhere, but will look at it after I've released this.

cloudfront/wellcomelibrary.org/edge-lambda/testRedirects.ts
```diff
@@ -179,11 +178,23 @@ const blogTestSet = {
...
   envs: {
-    stage: 'https://stage.wellcomelibrary.org',
-    prod: 'https://wellcomelibrary.org',
+    stage: 'https://blog.stage.wellcomelibrary.org/',
+    prod: 'https://blog.wellcomelibrary.org/',
+  },
+  checkResponse: checkMatchingUrl,
+};
+
+const archiveTestSet = {
```
kenoir, line 190: You could add these to the buildkite steps in a later PR.

jamesgorrie: Indeed that's the plan!

cloudfront/wellcomelibrary.org/edge-lambda/testRedirects.ts
```diff
@@ -179,11 +178,23 @@ const blogTestSet = {
...
   displayName: 'Apex (Content management) pages',
   fileLocation: staticRedirectFileLocation,
-  fileHostPrefix: staticRedirectsHost,
-  headers: staticRedirectHeaders,
+  fileHostPrefix: 'blog.wellcomelibrary.org',
+  headers: ['sourceUrl', 'targetUrl'],
   envs: {
-    stage: 'https://stage.wellcomelibrary.org',
-    prod: 'https://wellcomelibrary.org',
+    stage: 'https://blog.stage.wellcomelibrary.org/',
```
kenoir, line 184: Why has this become `blog.`?

kenoir (approved):
Just one question about the tests, but looks good otherwise.

cloudfront/wellcomelibrary.org/edge-lambda/testRedirects.ts
jamesgorrie: No idea what I was trying to do here, reverted.

PR #198, opened 2021-06-15
add redirections for collections/browse

jamesgorrie:
## What's changing and why?
Adding redirectes from `wl.org/collections/browse/*` => `wc.org/collections`.

Examples were taken from analytics.

## `terraform plan` diff
Coming soon once code is reviewed.


kenoir (approved):
✨

PR #202, opened 2021-06-18
Add new AWS account for DAM Prototype

kenoir:
## What's changing and why?

In order to support the DAM prototyping work (getting the storage service set up from scratch) this change adds a new AWS account to ringfence those resources.

### DIFF
--- a/accounts/dam_prototype/vpcs.tf
+++ b/accounts/dam_prototype/vpcs.tf
@@ -1,0 +1,6 @@
+module "dam_prototype_vpc" {
+  source = "../modules/vpc"
+
+  vpc_name   = "dam-prototype"
+  cidr_block = "172.31.0.0/16"
+}
--- a/accounts/identity/account.tf
+++ b/accounts/identity/account.tf
@@ -1,10 +1,10 @@
 module "identity_account" {
   source = "../modules/account/aws"
 
-  prefix = "dam_prototype"
+  prefix = "identity"
 
   principals = [
     local.account_principals["platform"],
-    local.account_principals["dam_prototype"],
+    local.account_principals["identity"],
   ]
 }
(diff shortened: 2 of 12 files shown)


alexwlchan:
Nitpick: the usual term is "dams", not "dam" = "Digital Asset Management System"

accounts/dam_prototype/vpcs.tf
```diff
@@ -0,0 +1,6 @@
+module "dam_prototype_vpc" {
+  source = "../modules/vpc"
+
+  vpc_name   = "dam-prototype"
+  cidr_block = "172.31.0.0/16"
+}
```
alexwlchan, line 6: I wonder: do we need this VPC?

I'm hoping to make a single Terraform configuration you can run `terraform plan/apply` in, and it spins up a copy of the storage service in a new account. The theory being that you could swap out the Terraform provider (it'll be `dams-admin`, but you could use your own) and it would all work for you.

It'll end up referring to the VPC module in this repo (pinned at a particular commit, for now), and using that VPC for the services. So I'm not sure I'll actually use this.

accounts/identity/account.tf
```diff
@@ -1,10 +1,10 @@
 module "identity_account" {
   source = "../modules/account/aws"
 
-  prefix = "identity"
+  prefix = "dam_prototype"
```
alexwlchan, line 4: 😱

accounts/platform/rolesets.tf
```diff
@@ -76,6 +81,7 @@ module "super_dev_roleset" {
     local.identity_account_roles["ci_role_arn"],
     local.digitisation_account_roles["ci_role_arn"],
     local.experience_account_roles["ci_role_arn"],
+    local.dam_prototype_account_roles["ci_role_arn"],
```
alexwlchan, line 84: Does CI need access to this account? I don't think BuildKite will be doing much here.

accounts/dam_prototype/vpcs.tf
kenoir: That's fine - i'm just following the new account pattern _just in case_.

accounts/platform/rolesets.tf
kenoir: Same as above. It's ok not to use it.

PR #204, opened 2021-06-22
redirect passthru redirects to wellcomecollection.org

jamesgorrie:
## What's changing and why?
Anything that passes though from wl.org will be redirected to wc.org. This will allow us to decommission the esrvers serving the apex domain.

This was the least amount of change to get what we want. There will be further tidy up PRs as with any far reaching redirect.

## `terraform plan` diff
no-op until lambdas are buildt

### DIFF
--- a/cloudfront/wellcomelibrary.org/edge-lambda/src/wellcomeLibraryPassthru.test.ts
+++ b/cloudfront/wellcomelibrary.org/edge-lambda/src/wellcomeLibraryPassthru.test.ts
@@ -47,4 +47,6 @@ test('redirects wellcomelibrary.org passthrus to wellcomecollection.org', async 
       { key: 'Access-Control-Allow-Origin', value: '*' },
     ],
   });
+
+  expect(originRequest.status).toEqual('302');
 });
--- a/cloudfront/wellcomelibrary.org/edge-lambda/src/wellcomeLibraryPassthru.test.ts
+++ b/cloudfront/wellcomelibrary.org/edge-lambda/src/wellcomeLibraryPassthru.test.ts
@@ -34,14 +34,17 @@ test('http requests are redirected to https', () => {
   );
 });
 
-test('rewrites the host header if it exists', async () => {
-  const request = testRequest('/', undefined, {
+test('redirects wellcomelibrary.org passthrus to wellcomecollection.org', async () => {
+  const request = testRequest('/unknown', undefined, {
     host: [{ key: 'host', value: 'www.wellcomelibrary.org' }],
   });
 
   const originRequest = await origin.requestHandler(request, {} as Context);
 
   expect(originRequest.headers).toStrictEqual({
-    host: [{ key: 'host', value: 'wellcomelibrary.org' }],
+    location: [{ key: 'Location', value: 'https://wellcomecollection.org/' }],
+    'access-control-allow-origin': [
+      { key: 'Access-Control-Allow-Origin', value: '*' },
+    ],
   });
 });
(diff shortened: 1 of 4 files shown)


cloudfront/wellcomelibrary.org/edge-lambda/src/wellcomeLibraryPassthru.test.ts
```diff
@@ -34,14 +34,17 @@ test('http requests are redirected to https', () => {
...
   });
 
   const originRequest = await origin.requestHandler(request, {} as Context);
 
   expect(originRequest.headers).toStrictEqual({
-    host: [{ key: 'host', value: 'wellcomelibrary.org' }],
+    location: [{ key: 'Location', value: 'https://wellcomecollection.org/' }],
+    'access-control-allow-origin': [
+      { key: 'Access-Control-Allow-Origin', value: '*' },
+    ],
```
alexwlchan, line 48: Do you want a test here that we return a 302 response? You could return Location headers on a 200 OK.

cloudfront/wellcomelibrary.org/edge-lambda/src/wellcomeLibraryPassthru.test.ts
```diff
@@ -34,14 +34,17 @@ test('http requests are redirected to https', () => {
   );
 });
 
-test('rewrites the host header if it exists', async () => {
-  const request = testRequest('/', undefined, {
+test('redirects wellcomelibrary.org passthrus to wellcomecollection.org', async () => {
+  const request = testRequest('/unknown', undefined, {
```
alexwlchan, line 38: What happens if I do hit `wellcomelibrary.org` after this update? Where will I be redirected then?

Is it worth keeping the existing test, and adding this behaviour as a new test?

cloudfront/wellcomelibrary.org/edge-lambda/src/wellcomeLibraryPassthru.test.ts
jamesgorrie: Nice idea, just did it.

cloudfront/wellcomelibrary.org/edge-lambda/src/wellcomeLibraryPassthru.test.ts
jamesgorrie: This test was in place for this bit of code:
https://github.com/wellcomecollection/platform-infrastructure/pull/204/files#diff-ce2a66225bb7a312c1db2c4434c46c33e11141de76621ea6a370bd33df51d45cL18. 

We've removed that so can't test it.

When hitting `/` you will still be redirected to `wc.org/` because we have no other redirect functionality for it. Does it feel like we need something specific for `/` as it's a special URL even though it just falls back to this?

alexwlchan: Ah, the change to request `/unknown` made this unclear – I thought that was changed because it didn't affect the `/` behaviour.

alexwlchan (approved):

cloudfront/wellcomelibrary.org/edge-lambda/src/wellcomeLibraryPassthru.test.ts
jamesgorrie: Yea, sorry I should have mentioned I repurposed the test for a completely different test.

PR #212, opened 2021-06-30
Make Fluentbit and API Gateway nginx images available in ECR Public

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5226 and https://github.com/wellcomecollection/platform/issues/5206. Supercedes/closes https://github.com/wellcomecollection/platform-infrastructure/pull/211.

## What's changing and why?

We now have ECR Public repositories for both `fluentbit` and `nginx_apigw`, in addition to the existing private repositories. The publishing scripts will publish to both repositories.

This will help somebody use the storage service in a non-Wellcome account – they can pull the images from ECR Public, rather than getting cross-account ECR permissions.

## `terraform plan` diff
Already applied.


images/nginx/loris.Dockerfile
```diff
@@ -1,3 +0,0 @@
-FROM nginx:1.15.8-alpine
-
-COPY loris.nginx.conf /etc/nginx/nginx.conf
```
alexwlchan, line 3: I've moved this into the Loris repo: https://github.com/wellcomecollection/loris-infrastructure/tree/main/nginx

We're never going to use it again so it shouldn't be here, but having it all in one place seems useful.

images/publish_fluentbit_image.sh
```diff
@@ -13,25 +16,46 @@ echo "*** WARNING! ***"
...
 echo ""
-echo "*** Publishing fluentbit ***"
+echo "*** Building fluentbit image ***"
 echo ""
 
-"$ROOT"/scripts/docker_run.py \
-      --dind --root -- \
-      wellcome/image_builder:23 \
-            --project=fluentbit \
-            --file=images/fluentbit/Dockerfile
```
alexwlchan, line 24: Trying to find where these Docker images are defined and what they do was a bit of a faff. Given how relatively simple the shell commands are, I opted to redo the scripts rather than shell out to some mystery images.

images/terraform/repo_pair/main.tf
```diff
@@ -0,0 +1,14 @@
+resource "aws_ecr_repository" "private" {
+  name = "${var.namespace}/${var.repo_name}"
+}
+
+resource "aws_ecrpublic_repository" "public" {
+  provider = aws.ecr_public
```
alexwlchan, line 6: In case you're wondering: ECR Public is another global resource, like IAM or CloudFront, and it lives in us-east-1.

images/publish_fluentbit_image.sh
kenoir: We almost never touch those images, and they (i think) are in the very unloved dockerfiles repo that i'd like to get rid of anyway.

kenoir (approved):

images/publish_fluentbit_image.sh
alexwlchan: They’re in the *history* of that repo, not even the latest commit. 😭

PR #213, opened 2021-07-02
Allow CI to get auth for ECR Public

alexwlchan:
## What's changing and why?

We need to run `aws ecr-public get-login-password` in CI so we can publish images to ECR Public. This extends that permission to all CI accounts. For https://github.com/wellcomecollection/platform/issues/5226

## `terraform plan` diff

Here's an example plan:

```
Terraform will perform the following actions:

  # module.catalogue_account.module.ci_role_policy.aws_iam_role_policy.ci_permissions will be updated in-place
  ~ resource "aws_iam_role_policy" "ci_permissions" {
        id     = "catalogue-ci:terraform-20200713113726102900000001"
        name   = "terraform-20200713113726102900000001"
      ~ policy = jsonencode(
          ~ {
              ~ Statement = [
                  ~ {
                      ~ Action   = [
                            # (15 unchanged elements hidden)
                            "ecr-public:InitiateLayerUpload",
                          + "ecr-public:GetAuthorizationToken",
                            "ecr-public:Get*",
                            # (3 unchanged elements hidden)
                        ]
                        # (3 unchanged elements hidden)
                    },
                    {
                        Action   = "s3:*"
                        Effect   = "Allow"
                        Resource = [
                            "arn:aws:s3:::wellcomecollection-catalogue-infra-delta/*",
                            "arn:aws:s3:::wellcomecollection-catalogue-infra-delta",
                        ]
                        Sid      = ""
                    },
                    # (7 unchanged elements hidden)
                ]
                # (1 unchanged element hidden)
            }
        )
        # (1 unchanged attribute hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```

PR #214, opened 2021-07-02
Allow the CI role to assume itself by default

kenoir:
## What's changing and why?

This change adds the ability for a CI role to assume itself by default. This is because we have multiple occasions where scripts in CI specify that they wish to assume the CI role to function correctly. In CI we must explicitly assume a role to grant permissions and this is by default the CI role for that account ... so it needs to be able to assume itself to function correctly.

It's not unreasonable for such scripts to do so, so make it the default rather than add exceptions.

**Note:** this change is already applied.


accounts/modules/role_policies/ci/policies.tf
```diff
@@ -17,6 +17,7 @@ data "aws_iam_policy_document" "ci_permissions" {
       "ecr-public:BatchCheckLayerAvailability",
       "ecr-public:Describe*",
       "ecr-public:Get*",
+      "ecr-public:GetAuthorizationToken",
```
kenoir, line 20: In order to get terraform to no-op I needed to add this permission - i'm not certain if it's been added manually or what the consequence will be if I remove it (so it gets added incidentally here).

alexwlchan: It's from https://github.com/wellcomecollection/platform-infrastructure/pull/213 – I need it to get publishing to ECR Public working in CI.

alexwlchan (approved):

PR #219, opened 2021-07-26
Logging alerting config scripts

jamesgorrie:
## What's changing and why?
This allows us to programatically set the alerting settings we have on the logging cluster. This felt useful in the sense of we should keep what the settings are in code. Unfortunately there is no way currently of setting more of this in terraform.

I've gone for having it as a script that is run as and when, rather than something that is automatically called when you create the EC deployment with TF as we will be changing this throughout the lifecycle of the deployment and don't want to have to recreate every time there is change.

Currently the only thing we can do is add the slack connector to certain rules.

## `terraform plan` diff
No-op, just scripts


logging/README.md
```diff
@@ -0,0 +1,3 @@
+# Logging
+
+## Scripts to manage the Elastic config
```
kenoir, line 3: Would be nice to add:
- What is each one for specifically?
- How (and when) do we run them?

logging/kibana.ts
```diff
@@ -0,0 +1,29 @@
+import fetch, { RequestInit } from "node-fetch";
+
+class KibanaClient {
```
kenoir, line 3: I'm surprised there isn't a Kibana client published by Elastic.

kenoir (approved):

logging/kibana.ts
jamesgorrie: So was I - I'm not sure if it's about stability, or just another library to maintain...

logging/README.md
jamesgorrie: Nice spot, definitely a half formed doc. Updated - what do you think?

PR #221, opened 2021-07-30
Add a common Lambda for alerting on non-empty SQS DLQs

alexwlchan:

This patch adds a Lambda function that can receive CloudWatch alarms about messages on an SQS DLQ, and send a notification to our alerts channel:

Screenshot 2021-07-30 at 16 44 45

The Lambda is copied to all three of our accounts (platform, catalogue, storage) and there's an alarm topic ARN for each account – so we don't have to deal with sending CloudWatch alarms across different AWS accounts.

**This means we'll start to get Slack alerts for DLQs from the storage/catalogue accounts, which we don't currently get.**

This Lambda function is much simpler than our old post-to-slack function – it's meant to handle DLQ alerts and DLQ alerts only, rather than lots of different types of message.

I wired this up to a test queue; to actually use it we'll need to update the Terraform of our assorted services.

I've also added some documentation to explain how this works: https://github.com/wellcomecollection/platform-infrastructure/blob/common-sqs-monitoring/monitoring/dlq_to_slack_alerts/README.md


critical/modules/secrets/distributed/main.tf
```diff
@@ -1,5 +1,5 @@
 module "secrets" {
-  source = "../secret"
+  source = "github.com/wellcomecollection/terraform-aws-secrets.git?ref=8bf3759af9e1a19731f28784953dd0a53a2f276f"
```
alexwlchan, line 2: This depends on https://github.com/wellcomecollection/terraform-aws-secrets/pull/4

jamesgorrie: Reviewed this - we can update this.

monitoring/dlq_to_slack_alerts/src/dlq_to_slack_alerts.py
```diff
@@ -0,0 +1,87 @@
...
+
+    print("Sending message %s" % json.dumps(slack_payload))
+
+    req = urllib.request.Request(
+        webhook_url,
+        data=json.dumps(slack_payload).encode("utf8"),
+        headers={"Content-Type": "application/json"},
+    )
+    resp = urllib.request.urlopen(req)
+    print(resp)
```
jamesgorrie, line 87: Is this just for debugging the lambda?

monitoring/terraform/modules/dlq_to_slack_alerts/lambda.tf
```diff
@@ -0,0 +1,43 @@
+data "archive_file" "dlq_to_slack_alerts" {
+  type        = "zip"
+  source_file = "${path.module}/../../../dlq_to_slack_alerts/src/dlq_to_slack_alerts.py"
```
jamesgorrie, line 3: Could we have a note on how we deploy this lambda?

jamesgorrie (commented):
Just wondering if we could have insdtructions on how to deploy the lambda?

critical/modules/secrets/distributed/main.tf
alexwlchan: Done.

monitoring/dlq_to_slack_alerts/src/dlq_to_slack_alerts.py
alexwlchan: Yup, I've tweaked it to catch a non-okay error and throw an error.

(Alerts on busted Lambdas also needs some work, which is the next PR.)

monitoring/terraform/modules/dlq_to_slack_alerts/lambda.tf
alexwlchan: Added – the Lambda gets automatically deployed when you apply Terraform.

jamesgorrie (approved):

PR #223, opened 2021-08-02
Further Terraform fixes for the SQS DLQ alerting Lambda

alexwlchan:

* Add default tags everywhere
* Make sure the Lambda is working in the platform account
* Use a per-account name for the SNS topic and Lambda so it's easier to debug

### DIFF
--- a/monitoring/terraform/stack/grafana/main.tf
+++ b/monitoring/terraform/stack/grafana/main.tf
@@ -66,7 +66,7 @@ module "task" {
 }
 
 module "service" {
-  source = "github.com/wellcomecollection/terraform-aws-ecs-service.git//service?ref=v1.4.0"
+  source = "./service"
 
   service_name = var.namespace
   cluster_arn  = var.cluster_arn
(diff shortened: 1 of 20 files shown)


monitoring/terraform/stack/grafana/main.tf
```diff
@@ -66,7 +66,7 @@ module "task" {
 }
 
 module "service" {
-  source = "github.com/wellcomecollection/terraform-aws-ecs-service.git//service?ref=v1.4.0"
```
kenoir, line 69: Why has the service moved internal to the module? I'm guessing the version used here is wildly out of date and upgrading is impractical?

kenoir (approved):

monitoring/terraform/stack/grafana/main.tf
alexwlchan: Pretty much, yeah. The actual motivation was tagging, explained here: https://github.com/wellcomecollection/platform-infrastructure/pull/223/files#diff-7d3d8ee04d5c20fcde3b70bb66fb6ecd3691fdd3bd00c6ecd82ef66d164e6b8cR61-R74

The Grafana service is very old and doesn't support tagging. We should upgrade it to a newer version, probably, but it works and I have no particular desire to fiddle with it.

PR #226, opened 2021-08-09
Increase client_max_body_size for experience nginx

jamieparkinson:
This is necessary for the Lighthouse server, which sometimes receives requests that are slightly larger than 1m (the default)

### DIFF
--- a/images/nginx/experience.nginx.conf
+++ b/images/nginx/experience.nginx.conf
@@ -12,6 +12,8 @@ http {
     gzip_min_length 860;
     gzip_proxied    any;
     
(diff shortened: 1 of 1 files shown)


images/nginx/experience.nginx.conf
```diff
@@ -11,6 +11,8 @@ http {
     gzip_types      text/css application/javascript application/json;
     gzip_min_length 860;
     gzip_proxied    any;
+    
+    client_max_body_size 2m;
```
alexwlchan, line 15: Throw in a comment explaining why we've increased it?

alexwlchan (approved):

PR #228, opened 2021-08-11
Tidy up the S3 IAM roles for publishing to S3 from CI

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5260

* Bin a bunch of roles that we only used when we had multiple repos of shared libraries, rather than one big library
* Give Buildkite complete permission to do stuff in the S3 releases bucket, rather than having to change permissions for each new library we add

This change is already applied.


kenoir (approved):
💃

PR #233, opened 2021-08-17
Add a new Lambda for getting alerts from failed Lambdas

alexwlchan:
This adds a new Lambda which sends us alerts about failed Lambda invocations. As with our other new Lambdas, it's deployed into every account where we run Lambdas:

Screenshot 2021-08-17 at 17 36 57


jamesgorrie (approved):
Real nice

PR #234, opened 2021-08-18
Remove all of the old Post to Slack infrastructure

alexwlchan:
This removes the old Post to Slack monolith, now we have the per-account topics and Lambdas set up. There may be more alerting we want to do here, but let's leave that for separate tickets – what we have currently is a like-for-like replacement for the old infra.

Closes https://github.com/wellcomecollection/platform/issues/5245


kenoir (approved):
🪓

PR #241, opened 2021-09-02
Add a monthly report for costs data in Slack

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5281

This adds a monthly report that tells us how much we're spending. It should fire on the second Tuesday of every month (i.e., next Tuesday, and then every month after that).

You probably want to start with the explanation in the README (https://github.com/wellcomecollection/platform-infrastructure/tree/costs-report/costs_report), then get to the code.


kenoir (approved):
👌

PR #246, opened 2021-09-08
Get ECS notifications when tasks aren't starting consistently

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5293

We tap into the CloudWatch Event stream for ECS events which show a task failing to start, and if we see a "consistently failing to start" in the service events, we warn in Slack. That's the theory, anyway. Unfortunately I'm unable to reproduce that error on demand, gah! But hey, I think this works, and the pieces are there for the next time we see this error.


kenoir (dismissed):

monitoring/slack_alerts/ecs_tasks_cant_start_alert/src/ecs_tasks_cant_start_alert.py
```diff
@@ -0,0 +1,101 @@
...
+    events shown in the "Events" tab of the ECS console.
+    """
+    client = sess.client("ecs")
+    resp = client.describe_services(cluster=cluster_arn, services=[service_name])
+
+    return resp["services"][0]["events"]
+
+
+@functools.cache
+def get_secret_string(sess, *, secret_id):
```
jamieparkinson, line 35: I vaguely recall that we do/did this somewhere already but it does feel like a package of weco utils and other stuff that we use all over the place on PyPI would be useful

jamieparkinson (approved):

monitoring/slack_alerts/ecs_tasks_cant_start_alert/src/ecs_tasks_cant_start_alert.py
alexwlchan: We do this all over the place, this function is very common.

We have done that in the past (see https://github.com/wellcomecollection/aws_utils); I've deliberately eschewed it here because it makes it significantly harder to deploy and debug your Lambdas.

jamieparkinson (approved):
🎉

PR #251, opened 2021-09-21
Only test our own redirects

jamieparkinson:
We were seeing flaky tests due to redirect latency from Internet Archive servers - this change means that we only test the redirects that we're doing ourselves


cloudfront/wellcomelibrary.org/edge-lambda/testRedirects.ts
```diff
@@ -75,9 +75,15 @@ async function testRedirects(env: EnvId, redirectTestSet: RedirectTestSet) {
         from: from,
       } as RedirectResult;
 
+      const axiosConfig: AxiosRequestConfig = {
+        maxRedirects: 0,
```
alexwlchan, line 79: Needs a comment about why we don't want to follow redirects.

alexwlchan (commented):
Looks good, just needs a comment.

kenoir (approved):

PR #253, opened 2021-09-29
Disable the experience CloudFront alarms

alexwlchan:
Make these alarms stop spamming our Slack channel and hiding real errors.


alicefuzier (approved):

alexwlchan:
Merging because it's the current state of the alarms, but we might re-enable them later.

PR #266, opened 2021-11-22
Standardise nginx images

jamieparkinson:
What this does:

- Builds all the Nginx images with the same base Dockerfile
- Renames `experience` to `frontend` to make intentions clearer
- Pushes all images to ECR Public (@alexwlchan is there a reason we were only doing it for apigw?)

### DIFF
--- a/images/nginx/README.md
+++ b/images/nginx/README.md
@@ -1,0 +1,5 @@
+# Nginx configurations
+
+These are the nginx configurations used across the platform. 
+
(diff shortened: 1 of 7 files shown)


alexwlchan:
> is there a reason we were only doing it for apigw?

We're publishing these ECR images to ECR Public because they're used in the storage service demo, and I wanted people to be able to pull images without us setting up cross-account ECR permissions. We didn't need any of the other images in ECR Public.

https://github.com/wellcomecollection/storage-service/blob/main/demo/terraform/demo_stack/stack.tf#L85-L100

alexwlchan (approved):

PR #275, opened 2022-01-06
Set up a CloudFront distro for search.wellcomelibrary.org

alexwlchan:
Actually pointing the DNS at the distro seems to send you into a redirect loop; I'm not sure why yet.


alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5056, but closing because it doesn't seem possible to actually set this up correctly.

PR #277, opened 2022-01-10
Create redirections for the known Encore URL patterns

alexwlchan:
This adds the redirects described in wellcomecollection/platform#5056; follows #276

It's not deployed anywhere yet (either stage or prod), but the business logic is there. We'll also get logs about any paths we aren't redirecting successfully, so we can add more later if necessary.

### DIFF
--- a/cloudfront/wellcomelibrary.org/edge-lambda/src/tests/wellcomeLibraryEncoreRedirect.test.ts
+++ b/cloudfront/wellcomelibrary.org/edge-lambda/src/tests/wellcomeLibraryEncoreRedirect.test.ts
@@ -36,7 +36,7 @@ const encoreTests = [
     },
   ],
   [
-    'single search page',
+    'single record page',
     {
       path: '/iii/encore/record/C__Rb2475299',
       qs: '',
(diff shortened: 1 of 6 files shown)


cloudfront/wellcomelibrary.org/edge-lambda/src/tests/wellcomeLibraryEncoreRedirect.test.ts
```diff
@@ -0,0 +1,114 @@
...
+import testRequest from './testEventRequest';
+import { Context } from 'aws-lambda';
+import { results, resultWithIdentifier } from './catalogueApiFixtures';
+import { expectedRedirect } from './testHelpers';
+import axios from 'axios';
+import { expect, jest, test } from '@jest/globals';
+import { CatalogueResultsList } from '../catalogueApi';
+
+jest.mock('axios');
+const mockedAxios = axios as jest.Mocked<typeof axios>;
```
jamieparkinson, line 11: Not one for this PR but I'm very impressed by https://mswjs.io/ for mocking HTTP requests, it's definitely better than using mocks like this

jamieparkinson (approved):

cloudfront/wellcomelibrary.org/edge-lambda/src/tests/wellcomeLibraryEncoreRedirect.test.ts
alexwlchan: As you can probably guess, this is copy/pasted from other tests. I'm going to leave it because I suspect this code is going to be frozen once we finish turning off wellcomelibrary.org, but good to know for other projects.

PR #281, opened 2022-02-04
Replace our sbt_wrapper Docker image with a smaller, faster one

alexwlchan:
Working with Paul this week, I've been acutely aware of how slow our Scala builds still are. I remember the bad old days of multi-hour builds, and while our current ~15m build times are fast*er*, they're still not *fast*.

All of our Scala builds invoke sbt through an `sbt_wrapper` image (previously defined in [our dockerfiles repo](https://github.com/wellcomecollection/dockerfiles/blob/main/sbt_wrapper/Dockerfile), which has become very stale). Since it's on the hot path, I decided to see if I could make it faster.

I've done this in two ways:

1.  The final image is smaller: <s>~1.15GB</s> ~770MB compared to ~2GB.

    Still not small, but smaller.  This is achieved by starting from a smaller base image and removing dependencies that are only used for building the image, not running it. Although EC2 networking is fast, it's not instant, and reducing the image size will reduce Docker image pull times.

2.  The sbt version is cached properly.

    If you look at our builds, you pretty regularly see this message:

        getting org.scala-sbt sbt 1.4.1  (this may take some time)...

    But the whole point of using a Docker image is to cache this sort of thing, so we don't have to install it on the fly!

    It turns out we were caching sbt in a slightly suboptimal way, and having to redownload it on every build host.  Now that's fixed, which is explained in the README, because it's quite fiddly.

This image is already deployed, and should be picked up on your next Scala build. ✨ 

Part of https://github.com/wellcomecollection/platform/issues/5370 and https://github.com/wellcomecollection/platform/issues/5359

### DIFF
--- a/images/dockerfiles/sbt_wrapper/Dockerfile
+++ b/images/dockerfiles/sbt_wrapper/Dockerfile
@@ -1,0 +1,22 @@
+FROM openjdk:11.0-jre-slim
+
+LABEL maintainer = "Wellcome Collection <dev@wellcomecollection.org>"
+LABEL description "A Docker image for running SBT"
+
+ENV SBT_VERSION 1.4.1
+ENV SBT_HOME /usr/local/sbt
+ENV PATH ${PATH}:${SBT_HOME}/bin
+
+COPY install_docker_compose.sh /
+RUN /install_docker_compose.sh
+
+COPY install_sbt.sh /
+RUN /install_sbt.sh
+
+VOLUME /repo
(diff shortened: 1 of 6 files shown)


alexwlchan:
> How do we roll this out safely?

It's been rolled out for a fortnight. 🤠 

I pushed a version under a different tag, and did some PR builds in pipeline with the tagged image. Then when it worked and PRs didn't break, I pushed it into the latest tag.

PR #282, opened 2022-02-08
Put a README in redirect buckets

jamieparkinson:
The buckets used for redirects look suspiciously like they do nothing, and so we're liable to delete them. This makes that less likely.


alexwlchan (changes requested):
Let's throw some default tags in the provider also?

alexwlchan (approved):
🎉

PR #288, opened 2022-03-16
Implement a lambda that posts Auth0 log stream alerts to slack

jamieparkinson:

### DIFF
--- a/monitoring/slack_alerts/auth0_log_stream_alert/src/auth0_log_stream_alert.py
+++ b/monitoring/slack_alerts/auth0_log_stream_alert/src/auth0_log_stream_alert.py
@@ -1,6 +1,13 @@
 """
 This is a Lambda which receives events from the Auth0 log stream rule in https://github.com/wellcomecollection/identity/blob/main/infra/scoped/auth0-logs.tf
-That file describes their schema.
+
+They look like:
+
+{
+  "environment": string,
+  "tenant_name": string
+  "log_id": string,
+}
 
 It sends alerts to Slack which link back to the log event in Auth0.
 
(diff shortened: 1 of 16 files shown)


monitoring/slack_alerts/auth0_log_stream_alert/src/auth0_log_stream_alert.py
```diff
@@ -0,0 +1,123 @@
...
+        ],
+    }
+
+    req = urllib.request.Request(
+        webhook_url,
+        data=json.dumps(slack_payload).encode("utf8"),
+        headers={"Content-Type": "application/json"},
+    )
+    resp = urllib.request.urlopen(req)
+    assert resp.status == 200, resp
```
paul-butcher, line 123: If this assert raises (and is logged as output), I think `resp` will be logged as its repr - e.g. '<http.client.HTTPResponse object at 0x107ae3580>'

`vars(resp)` would be a quick way to show everything it knows, but it won't have read any data by now (if that's important).

However, I'm not sure this is likely to raise AssertionError, as it would only happen when the request is successful, but not 200 (see above), which I don't think web hooks do.

monitoring/slack_alerts/auth0_log_stream_alert/src/auth0_log_stream_alert.py
```diff
@@ -0,0 +1,123 @@
...
+            }
+        ],
+    }
+
+    req = urllib.request.Request(
+        webhook_url,
+        data=json.dumps(slack_payload).encode("utf8"),
+        headers={"Content-Type": "application/json"},
+    )
+    resp = urllib.request.urlopen(req)
```
paul-butcher, line 122: url open raises on an error response, and will end up logging the Error as something unhelpful like urllib.error.HTTPError: HTTP Error 400 - Bad Request.

Slack tries to return some useful info in the response body for failed requests, so it would be better to catch this and try to read the body.

```suggestion
    from urllib.error import HTTPError

    try:
      resp = urllib.request.urlopen(req)
    except HTTPError as err:
      raise Exception(f"{err} - {err.read()}")
```

paul-butcher (approved):
Just one small improvement.

PR #304, opened 2022-06-13
Remove some mirrored ECR repos we no longer need

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5545

This removes a bunch of repos and mirrored images where we're now pulling from the ECR Public-maintained mirrors of the Docker Hub library images (or we should be), and so we can delete our copies.

This only leaves a handful of mirrored images:

* LocalStack, which we use for simulating AWS services in tests
* scality/s3server and zenko/cloudserver, which we use for simulating S3 and I'd like to eventually replace with LocalStack (see https://github.com/wellcomecollection/platform/issues/5547)

Plus a bunch of Wellcome-specific images, some of which are probably redundant (e.g. `image_builder` has largely been replaced by inline shell scripts or similar).

I haven't applied this yet in case somebody wants to save something, but I think all of these can be safely deleted.

### DIFF
--- a/images/publish_mirrored_images.py
+++ b/images/publish_mirrored_images.py
@@ -40,20 +40,9 @@ PLATFORM_ROLE_ARN = "arn:aws:iam::760097843905:role/platform-developer"
 
 # These repositories must be provisioned in terraform/ecr.tf
 IMAGE_TAGS = [
-    "amazon/aws-cli",
-    "amazon/dynamodb-local",
-    "hashicorp/terraform:light",
-    "library/mysql:5.6",
     "localstack/localstack",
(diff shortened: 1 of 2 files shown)


alexwlchan:
Note: my long-term thinking is that we'd:

* Have our images defined in the `images` repo of this repository, and published from here (or maybe broken out as a standard docker-infrastructure repository)
* Have a simpler shell script you can invoke to mirror a single image to ECR, e.g.

    ```console
    $ ./mirror_docker_image_to_ecr.sh "localstack/localstack:0.16.0"
    ```

    and we do away with the big Python script.

jamieparkinson (approved):

PR #309, opened 2022-06-16
Remove the template provider from the monitoring stack

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5557

I think this should allow @melanierogan to use Terraform on her M1 MacBook without getting warnings about an unavailable provider. 🤞 I've had to inline the module which defines our Grafana instance, which is v old – tbh we should probably move this to managed Grafana or similar, but it's never been high enough priority.


melanierogan (approved):
This looks good to me, I think the true test will be if it allows me to terraform on my laptop but also lets everyone else continue as is. Once in i'll rebase and try it

alexwlchan:
If it doesn't Terraform cleanly, feel free to rope me in to have another look, we should get to a point where you can Terraform nicely.

PR #318, opened 2022-08-01
Provide more information in the Auth0 Slack alerts

alexwlchan:
We don't send the description from the log event to Slack because it might contain PII, but there is a "log event type code" that we can map to a human-readable description.

This should make triage of events slightly easier, by allowing us to make an initial assessment before we open the dashboard.

Before:

Screenshot 2022-08-01 at 09 02 33

After:

Screenshot 2022-08-01 at 09 02 35

The event descriptions are populated from a fixed list of strings in a JSON file (taken from the Auth0 website), so we're not going to expose PII from the log event descriptions or something.


jamieparkinson (approved):
Good idea

PR #320, opened 2022-08-08
More small improvements to the Auth0 alerts

alexwlchan:
* Ignore signup errors when password contains user info
* Ignore a simple 'Unauthorized' error on login
* Log the endpoint which is being rate limited if somebody hits the rate limits


monitoring/slack_alerts/auth0_log_stream_alert/src/auth0_log_stream_alert.py
```diff
@@ -143,6 +145,33 @@ def should_alert_for_event(log_event):
...
+    we know it's safe.
+    """
+    # These keys should match those set in the event transform rule
+    # https://github.com/wellcomecollection/identity/blob/main/infra/scoped/auth0-logs.tf#L55
+    log_event_type = log_event["log_event_type"]
+    log_description = log_event["log_description"]
+
+    # If we're hitting API rate limits, we can log the name of the
+    # endpoint that's being limited without giving anything away.
+    if log_event_type == "api_limit" and re.match(
```
paul-butcher, line 167: I'd prefer this, but only marginally
```suggestion
    return log_event_type == "api_limit" and re.match(
```

paul-butcher (approved):
Looks fine, just one small style suggestion you can take or leave.

monitoring/slack_alerts/auth0_log_stream_alert/src/auth0_log_stream_alert.py
alexwlchan: I'm going to leave this as-is; the intention was that over time we might add other `if` branches for cases where it's safe to let the description through.

PR #324, opened 2022-08-22
Add redirects for alpha.wc.org and blog.wc.org

alexwlchan:
## What's changing and why?

Adding redirects for alpha.wc.org and blog.wc.org, which are both defunct subdomains we want to be redirecting to the new site.

For https://github.com/wellcomecollection/wellcomecollection.org/issues/8273; for https://github.com/wellcomecollection/platform/issues/4462

## `terraform plan` diff

Already applied.

### DIFF
--- a/cloudfront/wellcomecollection.org/terraform/main.tf
+++ b/cloudfront/wellcomecollection.org/terraform/main.tf
@@ -22,6 +22,8 @@ locals {
 
   # Subdomains that should be redirected to wellcomecollection.org
   redirect_subdomains_to_apex = [
+    "alpha.wellcomecollection.org",
+    "blog.wellcomecollection.org",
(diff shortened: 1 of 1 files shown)


alexwlchan:
This is applied, I want to test it before we merge it. It takes ~5m for new DNS records to propagate (if caches use the configured TTL).

alexwlchan:
There's been plenty of time for this DNS change to apply, and it hasn't. Redirects from www.wc.org also seem flaky for me, even though I haven't touched them.

HMM.

alexwlchan:
So specifically, it looks like redirects don't work for https:// URLs, but they do for http:// URLs.

I'm not sure how this has ever worked, and right now I don't care enough to fix it properly. I'll open a separate ticket for that.

PR #337, opened 2022-10-12
Get the reporting cluster in Terraform

alexwlchan:
Closes https://github.com/wellcomecollection/reporting/issues/82

For some reason it's failing with

```
ec_deployment.reporting: Modifying... [id=d3f9c38fe7134d44b3ec7752d86b5e98]
╷
│ Error: failed updating deployment: 1 error occurred:
│ 	* api error: deployments.elasticsearch.node_type.unsupported_version: The node_type field in the Elasticsearch topology is not supported in this version of elasticsearch.
│
│
│
│   with ec_deployment.reporting,
│   on elastic_reporting_cluster.tf line 79, in resource "ec_deployment" "reporting":
│   79: resource "ec_deployment" "reporting" {
```

Something weird is happening because we have data nodes and ML nodes. 🤷‍♀️


alexwlchan:
The tricky part was the ML node, which Jamie has said was just created experimentally – so I think if I delete that I might be able to get this working.

alexwlchan:
I've deleted the ML node, so this should be equivalent to the logging cluster – but for some reason Elasticsearch is still giving me the same error. I’m a bit stuck. 😕

alexwlchan:
It seems to be planning/applying cleanly; the trick is to make Terraform think there's nothing to change – it wanted to change whitespace in the user settings YAML, which we don't care about. Fixing that, it doesn't touch the reporting cluster and so it plans/applies cleanly.

jamieparkinson (approved):

PR #345, opened 2022-12-02
Improve Terraform for Elasticsearch clusters

jamieparkinson:
This is already applied as it was a bit of a trial and error process - it moves the logging cluster's configuration out of a script and into Terraform, and it fixes the reporting cluster Terraform so that it can be used.

I believe that the issues we'd been seeing [previously](https://github.com/wellcomecollection/platform-infrastructure/pull/337#issuecomment-1305270567) were caused by us (a) not configuring our topologies quite right (see ["Note on Elasticsearch topology IDs"](https://registry.terraform.io/providers/elastic/ec/latest/docs/resources/ec_deployment#topology)) and (b) the leftover `0.1.0-beta` version of the provider finding its way into places it shouldn't be!

### DIFF
--- a/critical/elastic_logging_cluster.tf
+++ b/critical/elastic_logging_cluster.tf
@@ -75,7 +75,7 @@ resource "ec_deployment" "logging" {
 
   elasticsearch {
     topology {
-      id = "hot_content"
+      id         = "hot_content"
       zone_count = 3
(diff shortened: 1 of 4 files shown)


alexwlchan (approved):
🔪 🐍

PR #351, opened 2022-12-16
Bump APM node size

jamieparkinson:
This matches @alexwlchan's changes in the console and is a good idea regardless


alexwlchan (approved):
Looks fine, although I didn't realise I was bumping it – I just picked whatever their default was.

PR #352, opened 2022-12-16
Configure the Kinesis stream with a cross-account log destination

jamieparkinson:
This makes it possible to send logs to Kinesis from any account, and also fixes the ESF configuration.


alexwlchan (approved):
🛴

PR #373, opened 2023-03-21
Get Slack alerts from the CloudFront logs for the API distribution

alexwlchan:
This is similar to what we're doing for the front-end – any errors will be posted in Slack along with the URL that caused the error. Most of this code is copied directly from the front-end repo.

For https://github.com/wellcomecollection/platform/issues/5671


cloudfront/api.wellcomecollection.org/send_slack_alert_for_5xx_errors.js
```diff
@@ -0,0 +1,339 @@
...
+ *      2022-05-18	14:11:09	/works/atry66dj/items
+ *
+ * This function emits objects so that downstream callers don't need
+ * to know about the structure of this file, e.g.
+ *
+ *      { date: '2022-05-18', time: '14:11:10', cs-uri-stem: '/search/works' }
+ *      { date: '2022-05-18', time: '14:11:09', cs-uri-stem: '/works/atry66dj/items' }
+ *
+ */
+async function findCloudFrontHitsFromLog(bucket, key) {
```
jamieparkinson, line 45: The implementation of this is strongly Python-developery 🐍

jamieparkinson (approved):

cloudfront/api.wellcomecollection.org/send_slack_alert_for_5xx_errors.js
alexwlchan: Well it was originally written in Snekk-script. 😜 

The only reason it's in JavaScript now is that I was putting it in the front-end repo where there's very little (any?) Python and it felt like introducing another language was creating an unnecessary complexity.

PR #382, opened 2023-04-25
Update Cloudfront to point iiif-test.wc.org at new DLCS

tomcrane:
## What's changing and why?

We have a prod configuration of new DDS and new DLCS, and we want to point the public iiif-test.wellcomecollection.org at it.

But it needs origins updating.


alexwlchan (approved):

alexwlchan:
Looks fine, do you need me to apply it or can you do it yourself?

donaldgray:
Thanks @alexwlchan, I think you'll need to apply. I was unable to plan as I can't assume `wellcomecollection-assume_role_hosted_zone_update` role.

PR #384, opened 2023-05-03
Add the Qualys EC2 scanning role for D&T

alexwlchan:
This is adding a standard IAM role to all our accounts, which allows the Qualys vulnerability scanning tool being used by the InfoSec team to scan EC2 instances in our accounts.

(I suspect they won't find much as we run very little vanilla EC2, but this role is pretty innocuous and it seems fine to me.)


alexwlchan:
(This role was written by somebody in D&T, so I'm gonna merge it rather than leave it lying around)

PR #390, opened 2023-06-05
SSO for modernised Grafana

jamieparkinson:
What's it do:
- Uses our standard ECS modules for the grafana instance
- Moves it to Fargate (which now supports EFS)
- Upgrades it to the latest version
- Makes users authenticate with Azure 
- Exposes it to the public internet
- Adds a privatelink endpoint for elastic cloud to the monitoring VPC (because I needed this for sending application logs)


monitoring/terraform/stack/security_groups.tf
```diff
@@ -1,5 +1,5 @@
 resource "aws_security_group" "service_egress_security_group" {
-  name        = "${var.namespace}_service_egress_security_group"
+  name        = "${var.namespace}-grafana_service_egress_security_group"
   description = "Allow traffic between services"
```
alexwlchan, line 3: this continues to be wrong 😢

monitoring/terraform/stack/service.tf
```diff
@@ -0,0 +1,118 @@
...
+module "task_definition" {
+  source = "github.com/wellcomecollection/terraform-aws-ecs-service//modules/task_definition?ref=v3.13.2"
+
+  task_name    = "${var.namespace}-grafana"
+  launch_types = ["FARGATE"]
+
+  // Grafana previously ran on a t2.small (1vCPU/2GB)
+  // That was probably too large but its CPU usage is very bursty
+  // and unfortunately Fargate is not burstable (t2s are), so
+  // we're trying this as a compromise between latency and cost
```
alexwlchan, line 84: A couple of thoughts:

* Would it be worth running our Grafana instance on Fargate Spot to help this compromise? It can afford to be interrupted occasionally.
* Digirati have tools to spin down certain environments (e.g. staging) outside office hours. I wonder if we should consider building something similar? We'd use it in other places too, e.g. the data science APIs, staging website, maybe bits of Archivematica. (Not for this PR, but future discussion.)

alexwlchan (commented):
Sweeeeeeet. This has been trundling along on legacy infra for way too long, thanks for mucking in and getting it upgraded to our newest modules.

One question (and maybe I just missed this): how do we connect the domain name monitoring.wc.org to our load balancer? Are we creating that DNS record in Terraform?

jamieparkinson:
Good point about the DNS - I think this predates our ability to manage DNS in Terraform, I'll see if I can import the record resource

monitoring/terraform/stack/service.tf
alexwlchan: I think it's just a single boolean parameter to the service module, easy to try it and roll back later.

PR #402, opened 2023-10-13
Update test/staging IIIF Cloudfront for IIIF Auth2

donaldgray:
## What's changing and why?

Added a new behaviour/origin for IIIF Auth v2 (test + staging env only to verify setup).

New origins are `dlcs.io/auth/v2/probe/*` and `dlcs.io/auth/v2/*`.

There was already a `dlcs_auth` behaviour which handled `/auth/*` paths (for IIIF auth v1). 

I added the new `dlcs_auth_2` / `/auth/v2/*` behaviour as a lower precedence to ensure it is evaluated before ^. This has resulted in some behaviours being re-ordered and the `terraform plan` being pretty large.

## `terraform plan` diff

I initially included `terraform plan` output but it was too large and I received an error: 

> Pull request creation failed. Validation failed: Body is too long (maximum is 65536 characters)


jamieparkinson (approved):
✨  looks good to me!

tomcrane (approved):

PR #404, opened 2023-10-24
Further updates to iiif-stage and iiif-test CloudFront config for iiif-auth2

donaldgray:
## What's changing and why?

This PR extends changes made via #402.

After testing deployed DLCS it became apparent that the CloudFront origin paths needed to be altered.

In addition to this the lambda@edge function needed some slight alterations to allow us to handle incoming paths correctly. It looks like the same lambda is used for all envs, this change should be safe to apply as there will be no prod requests that match `/^\/auth\/v2\/.+/` as this is a new path.


jamieparkinson (approved):
lgtm - I'll apply changes after it's merged?

PR #413, opened 2024-01-26
Remove healthcheck endpoints from nginx configuration for frontend

kenoir:
## What's changing and why?

This change removes the healthcheck endpoints from the frontend and frontend_identity docker images so these can now reach the app containers which serve there own healthchecks.

We also update the publish scripts to be a bit neater and use functions.

See https://github.com/wellcomecollection/wellcomecollection.org/issues/10545 for the motivation for this change.

In order to use this change, we should:

- [x] Build new versions of these container images, these will get pushed [ECR in the platform account](https://eu-west-1.console.aws.amazon.com/ecr/private-registry/repositories?region=eu-west-1).
  - `760097843905.dkr.ecr.eu-west-1.amazonaws.com/uk.ac.wellcome/nginx_frontend:b809c6ef4363ff0cbac8da7ff5e80d865cfcd008`
  - `760097843905.dkr.ecr.eu-west-1.amazonaws.com/uk.ac.wellcome/nginx_frontend_identity:b809c6ef4363ff0cbac8da7ff5e80d865cfcd008`
- [ ] Update the consuming services reference to the build images, for [the content service and identity (frontend)](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/infrastructure/modules/service/main.tf#L43), (content uses default variables for tag, identity overrides these).
- [ ] Optionally: remove the default references to the old container images from the [terraform-aws-ecs-service](https://github.com/wellcomecollection/terraform-aws-ecs-service/blob/main/modules/nginx/frontend/variables.tf#L23)
  - https://github.com/wellcomecollection/terraform-aws-ecs-service/pull/52

## How to test

- [ ] Update the staging environment for the content and identity services to use the new container image after ensuring the services serve healthcheck endpoints, needs: 
  - https://github.com/wellcomecollection/wellcomecollection.org/pull/10617
  - https://github.com/wellcomecollection/wellcomecollection.org/pull/10612

## How can we measure success?

Healthchecks that really test if the application is running protect us from broken deployments going out, and ensure the application container is really serving requests when marked as healthy.

### DIFF
--- a/images/dockerfiles/publish_nginx_images.sh
+++ b/images/dockerfiles/publish_nginx_images.sh
@@ -5,7 +5,6 @@ set -o errexit
 
 AWS_ACCOUNT_ID=760097843905
 AWS_PROFILE=platform-developer
-DEV_ROLE_ARN="arn:aws:iam::$AWS_ACCOUNT_ID:role/platform-developer"
 ROOT=$(git rev-parse --show-toplevel)
 CURRENT_COMMIT=$(git rev-parse HEAD)
 
--- a/images/dockerfiles/publish_nginx_images.sh
+++ b/images/dockerfiles/publish_nginx_images.sh
@@ -1,71 +1,96 @@
 #!/usr/bin/env bash
 
 set -o nounset
+set -o errexit
 
-DEV_ROLE_ARN="arn:aws:iam::760097843905:role/platform-developer"
+AWS_ACCOUNT_ID=760097843905
+AWS_PROFILE=platform-developer
+DEV_ROLE_ARN="arn:aws:iam::$AWS_ACCOUNT_ID:role/platform-developer"
 ROOT=$(git rev-parse --show-toplevel)
+CURRENT_COMMIT=$(git rev-parse HEAD)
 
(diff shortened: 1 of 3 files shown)


jamieparkinson (approved):

images/dockerfiles/publish_nginx_images.sh
```diff
@@ -1,71 +1,97 @@
 #!/usr/bin/env bash
 
 set -o nounset
+set -o errexit
 
-DEV_ROLE_ARN="arn:aws:iam::760097843905:role/platform-developer"
+AWS_ACCOUNT_ID=760097843905
+AWS_PROFILE=platform-developer
+DEV_ROLE_ARN="arn:aws:iam::$AWS_ACCOUNT_ID:role/platform-developer"
```
agnesgaroux, line 8: Could you use `$AWS_PROFILE` in the DEV_ROLE_ARN instead of `platform-developer`

kenoir: I think `DEV_ROLE_ARN` is now unused, so i'll remove that line entirely.

PR #439, opened 2024-06-13
Remove Grafana monitoring service (#5750)

StepanBrychta:
## What's changing and why?

https://github.com/wellcomecollection/platform/issues/5750

This removes all infrastructure related to the Grafana monitoring service.

## `terraform plan` diff

GitHub doesn't let me paste the full diff here because it exceeds the maximum allowed length, but this destroys 34 resources, mostly in the `monitoring/terraform/stack/` folder.


images/terraform/ecr.tf
```diff
@@ -49,19 +49,6 @@ module "ecr_nginx_frontend_identity" {
...
-  namespace = local.namespace
-  repo_name = "nginx_grafana"
-
-  description = "An nginx image for reverse proxying Grafana"
-
-  providers = {
-    aws.ecr_public = aws.ecr_public
-  }
-}
-
```
StepanBrychta, line 64: I'm not really sure what this was for but I didn't find any references to it and I suspect it wasn't used, so hopefully it's safe to remove.

kenoir: This creates both a public and private ECR repo. I believe it is safe to remove.

kenoir (approved):
This looks good to me. It will need a DNS update to remove the CNAME record for monitoring.wellcomecollection.org that we can talk through after this change is made.

Issue #443, opened 2024-08-06
Increase memory for send_slack_alert_for_5xx_errors lambda


agnesgaroux:
This lambda reads log files, looking for interesting errors to send as Slack alert
It's been itself erroring 🙃 
See https://wellcome.slack.com/archives/CQ720BG02/p1722872321976499

This has the potential to mask crucial errors if the lambda shuts down in the middle of reading a log file and fails to alert us

FIX
Increase the memory to 256

PR #444, opened 2024-08-06
increased memory from 128 (default) to 256

georgiaewhitney:
## What's changing and why?

Memory increased from 128 (default var) to 256 as per #443 

## `terraform plan` diff
```
Terraform used the selected providers to generate the following execution plan. Resource actions are
indicated with the following symbols:
  ~ update in-place
-/+ destroy and then create replacement

Terraform will perform the following actions:

  # aws_iam_role_policy.allow_s3_read must be replaced
-/+ resource "aws_iam_role_policy" "allow_s3_read" {
      ~ id          = "lambda_send_slack_alert_for_5xx_errors:terraform-20230321122311397500000001" -> (known after apply)
      ~ name        = "terraform-20230321122311397500000001" -> (known after apply)
      - name_prefix = "terraform-" -> null # forces replacement
        # (2 unchanged attributes hidden)
    }

  # module.slack_alerts_for_5xx.aws_lambda_function.lambda_function will be updated in-place
  ~ resource "aws_lambda_function" "lambda_function" {
        id                             = "send_slack_alert_for_5xx_errors"
      ~ last_modified                  = "2024-07-19T08:22:17.000+0000" -> (known after apply)
      ~ memory_size                    = 128 -> 256
      ~ source_code_hash               = "jsduuStdqBOd39Sw6YQfq1KAsEy5OzA2f4gZUEfxUdI=" -> "Muyhd941OhW1yIHHSbtKNyrHywvymTj0LntBt9NxVsE="
        tags                           = {}
        # (25 unchanged attributes hidden)

        # (4 unchanged blocks hidden)
    }

Plan: 1 to add, 1 to change, 1 to destroy.
```

### DIFF
--- a/cloudfront/api.wellcomecollection.org/.terraform.lock.hcl
+++ b/cloudfront/api.wellcomecollection.org/.terraform.lock.hcl
@@ -4,6 +4,7 @@
 provider "registry.terraform.io/hashicorp/archive" {
   version = "2.3.0"
   hashes = [
+    "h1:NaDbOqAcA9d8DiAS5/6+5smXwN3/+twJGb3QRiz6pNw=",
     "h1:pTPG9Kf1Qg2aPsZLXDa6OvLqsEXaMrKnp0Z4Q/TIBPA=",
     "zh:0869128d13abe12b297b0cd13b8767f10d6bf047f5afc4215615aabc39c2eb4f",
     "zh:481ed837d63ba3aa45dd8736da83e911e3509dee0e7961bf5c00ed2644f807b3",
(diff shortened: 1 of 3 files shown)


StepanBrychta (commented):
In the current setup, the `platform-infrastructure/cloudfront/api.wellcomecollection.org/lambda/` folder is a module which can be used as a blueprint to define specific lambda functions. The module defines several variables in `variables.tf` to keep things configurable, including the `memory_size` variable.

This change bypasses the variable and hardcodes a specific memory value into the module itself, meaning that all lambdas making use of the module will have a hardcoded memory size of 256 MB.

I think a better approach would be to keep the module as is and instead make use of the variable by setting `memory_size = 256` in the lambda function itself (defined in `platform-infrastructure/cloudfront/api.wellcomecollection.org/send_slack_alert_for_5xx_errors.tf`) to ensure that only this one lambda function is affected by this change.

(This probably won't make a difference in the `terraform plan` because it looks like there is currently only one lambda function using the module, but I think it would still be a good idea to keep the memory size configurable for future lambdas.)

georgiaewhitney:
Thanks for this, @StepanBrychta! This is really well explained and the approach you've suggested makes a lot of sense - I'll make these changes.

StepanBrychta (approved):
Looks good!

PR #458, opened 2024-12-09
up runtime and update deploy script for sdk v3

agnesgaroux:
## What does this change?

Part of https://github.com/wellcomecollection/platform/issues/5837

## How to test

???
The deploy script works alright for the invalidation lambdas

## How can we measure success?

`dlcs_path_rewrite` is on node20 and images are being served

## Have we considered potential risks?

We can't get any dlcs images on weco.org 😱

### DIFF
--- a/cloudfront/iiif.wellcomecollection.org/edge-lambda/Dockerfile
+++ b/cloudfront/iiif.wellcomecollection.org/edge-lambda/Dockerfile
@@ -1,4 +1,4 @@
-FROM public.ecr.aws/docker/library/node:16-slim
+FROM public.ecr.aws/docker/library/node:20-slim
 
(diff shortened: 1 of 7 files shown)


agnesgaroux:
Steps
- merge this
- it should create a new version of the `dlcs_path_rewrite` lambda (v32)
- set the runtime of this lambda to node20
- run the tf so that stage uses the latest (v32) while still  leaving prod on v31
- check it works in stage
- set `dlcs_path_rewrite_arn_prod` to `"${local.dlcs_path_rewrite_arn}:32"`

kenoir (approved):
The steps make sense, and give you an opportunity to check the prod lambda version remains unchanged when deploying to stage 👍

